PR #24, opened 2016-03-16
Small fix for nightly rust

kaksmet:


oli-obk:
this [has been fixed](https://github.com/rust-lang/rust/pull/32285) and should be available in tomorrows nightly. The commit can then be reverted.

PR #28, opened 2016-03-19
Fix compile error on Rust 1.9.0 nightly

ruud-v-a:


kaksmet:
This was caused by a [regression](https://github.com/rust-lang/rust/pull/32285) in nightly rust.
If you update to today's nightly it will compile again without this patch.

PR #58, opened 2016-11-20
Adds UpsamplerGeneric

### LINKED ISSUE #48: Support more chroma subsampling ratios
Implementing a generic integer upsampler will allow us to support most chroma subsampling ratios.

iamrohit7:
Implements a Generic Upsampler with support for integer sampling ratios. Fixes #48

### DIFF
--- a/src/upsampler.rs
+++ b/src/upsampler.rs
@@ -3,6 +3,7 @@ use parser::Component;
 
 pub struct Upsampler {
     components: Vec<UpsamplerComponent>,
+    line_buffer_size: usize
 }
 
 struct UpsamplerComponent {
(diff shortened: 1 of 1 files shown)


kaksmet:
When I try to decode the image in servo/servo#13164 using your branch it panics with

`thread '<unnamed>' panicked at 'index out of bounds: the len is 2561 but the index is 2561', src/upsampler.rs:241`.

It would be great to have some tests for this code. Maybe converting some of the images in `tests/reftest/images/mozilla` to use 3x sampling factor using `cjpeg`'s `-sample` option?

iamrohit7:
The tests are still failing, but the visual result looks fine. I haven't included the failing tests.

src/upsampler.rs
```diff
@@ -34,12 +35,14 @@ impl Upsampler {
...
             components: upsampler_components,
+            max_horizontal_sampling_factor: h_max
         })
     }
 
     pub fn upsample_and_interleave_row(&self, component_data: &[Vec<u8>], row: usize, output_width: usize, output: &mut [u8]) {
         let component_count = component_data.len();
-        let mut line_buffer = vec![0u8; output_width + 1];
+        let max_component_width = self.components.iter().map(|c| c.width).max().unwrap();
+        let mut line_buffer = vec![0u8; max_component_width * self.max_horizontal_sampling_factor as usize];
```
kaksmet: Here I would like to cache the result of `components.iter().map(|c| c.size.width as usize * c.horizontal_sampling_factor as usize).max().unwrap()` in Upsampler and use that for line_buffer's size instead.

src/upsampler.rs
```diff
@@ -214,3 +225,25 @@ impl Upsample for UpsamplerH2V2 {
...
+    // Uses nearest neighbor sampling
+    fn upsample_row(&self,
+                    input: &[u8],
+                    input_width: usize,
+                    _input_height: usize,
+                    row_stride: usize,
+                    row: usize,
+                    _output_width: usize,
+                    output: &mut [u8]) {
+        let mut index: usize = 0;
```
kaksmet: Remove the explicit type from index

PR #60, opened 2016-12-21
Bump Rayon to 0.6

bholley:
Servo needs to update to rayon 0.6, and jpeg-decoder is currently the only dependency it has that currently requires 0.5. Would you mind doing a quick release to unblock us?

CC @nikomatsakis @metajack @larsbergstrom @emilio @SimonSapin


bholley:
Thanks! Pulling this in: https://github.com/servo/servo/pull/14709

PR #73, opened 2017-05-31
Use rayon's `par_chunks_mut`

cuviper:
In `compute_image_parallel`, there's no need to collect `chunks_mut`
into a vector when rayon can iterate `par_chunks_mut` directly.  The
non-rayon version also doesn't need an intermediate vector.


kaksmet:
Thanks!

PR #82, opened 2017-10-23
Needless clone and pass by value

luisbg:
spectral_selection isn't consumed by decode_block, no need to clone it.
Also, no need to pass it by value, we can pass it by referene and avoid
an allocation.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -460,7 +460,7 @@ impl<R: Read> Decoder<R> {
                                          &mut huffman,
                                          self.dc_huffman_tables[scan.dc_table_indices[i]].as_ref(),
                                          self.ac_huffman_tables[scan.ac_table_indices[i]].as_ref(),
-                                         scan.spectral_selection.clone(),
+                                         &scan.spectral_selection,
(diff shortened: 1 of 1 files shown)


kaksmet:
Thanks

philipc:
Maybe I'm missing something, but isn't a `Range<u8>` simply two bytes, so cloning it should be a simple move instruction? And I would expect that passing it by value is cheaper than needing to dereference in `decode_block`. The only reason `Range` doesn't derive `Copy` is because there's a bad interaction with its `Iterator` implementation, but cloning is still cheap.

Additionally, the call to `decode_block_successive_approximation` still uses a clone.

kaksmet:
In practice cloning it or passing it by reference won't matter here, but since @philipc pointed out that this patch is incomplete, I will revert it.

luisbg:
Sorry I didn't answer sooner. I'm at a conference.

I will complete the patch in the next few days.

Thank you very much for the review 😃

luisbg:
Looks like the code should remain as it is.

Sorry for the trouble.

PR #87, opened 2018-06-08
Support WebAssembly and asm.js

### LINKED ISSUE #75: Support emscripten targets
At the moment a worker thread is used while decoding. However the emscripten targets (asm.js and WebAssembly) don't support threads, so a synchronous alternative implementation should be used for those targets. The worker thread could possibly just be executed on the main thread for example.

On a side note: rayon is used in other places, but that works just fine with emscripten targets, so supporting emscripten isn't all too big of a change.

CryZe:
This adds support for both WebAssembly and asm.js by abstracting the worker thread into an ImmediateWorker and a ThreadWorker which uses the ImmediateWorker. The ThreadWorker is exposed as the worker to be used on every platform other than WebAssembly and asm.js. There the ImmediateWorker itself is exposed instead, because threading isn't supported on those platforms just yet.

Resolves #75


kaksmet:
Looks good, thanks!
I'll try to get a new version published today or tomorrow.

PR #94, opened 2019-03-19
allow zero lengths, which matters for DHT

FauxFaux:
I have a `.jpg` produced by the camera app on a smartphone, which contains zero-length `DHT` segments.

I believe this is against the spec, B.2.4.2 specifies that a DHT declaration "defines one or more Huffman tables", from which I'm inferring that a packet of length zero (declaring zero specifications) is invalid.

All of the existing code seems to cope with empty tags, including the existing DHT implementation: if there are no updates, then it doesn't apply any updates. I'm hesitant to optimise the code for this illegal, presumably uncommon case (I've never seen it before).

This PR allows a length of `0` (i.e. `2` including the length), and updates the error handling inside the `SOS` parser to explicitly error in this case. The handling in everything else looks okay to me.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -85,7 +85,7 @@ fn read_length<R: Read>(reader: &mut R, marker: Marker) -> Result<usize> {
     // length is including itself.
     let length = reader.read_u16::<BigEndian>()? as usize;
 
-    if length <= 2 {
+    if length < 2 {
         return Err(Error::Format(format!("encountered {:?} with invalid length {}", marker, length)));
     }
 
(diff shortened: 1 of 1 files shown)


FauxFaux:
Here is the offending image, in a `.tar.gz` so github doesn't mess with it, under the "sure, just use it" license.

[broken.jpeg.tar.gz](https://github.com/kaksmet/jpeg-decoder/files/2985381/broken.jpeg.tar.gz)

kaksmet:
Looks good, thanks!

My stance on matters like this (spec ambiguity/violations) is that if there exists software/hardware in the wild producing such images and libjpeg(-turbo), being the reference implementation, correctly decodes them, then jpeg-decoder should also be able to correctly decode them.

PR #97, opened 2019-08-04
Remove byteorder-dependency

lukaslueg:
This removes the `byteorder`-dependency, which is only ever really used for a single big-endian conversion. Since Rust 1.32, we can use the functions defined on the primitive to do that. IMHO dropping a dependency is worth the change.


lovasoa:
The minimum rust version required to compile this crate was recently bumped from 1.24 to 1.28. 
@kaksmet : what is your opinion about moving it further it to 1.32 ?

PR #98, opened 2019-08-08
UnexpectedEof: failed to fill whole buffer

spl:
This is a bug report even though it is a PR. I'm using the PR to include an image that fails with:

```
thread 'reftest::reftest' panicked at 'failed to decode file:
"tests/reftest/images/curry.jpg":
Io(Custom { kind: UnexpectedEof, error: StringError("failed to fill whole buffer") })'
```

I have multiple images that fail with the same error, but I figured it would be easiest to start with one.


kaksmet:
Decoding the file with `djpeg` prints `Premature end of JPEG file`, meaning it has been truncated.
`jpeg-decoder` doesn't try to recover from errors produced by corrupt files.

spl:
Thanks. Unfortunately, this is the state of my files. Does that effectively mean I can't use `jpeg-decoder` (and thus also can't use the `resize_image` built-in function of [Zola](https://github.com/getzola/zola) which uses `jpeg-decoder` by transitivity)? Do you have any future plans to deal with such files in `jpeg-decoder`? Is there a way I can repair them?

fintelia:
It isn't clear to me from this proposal what this crate could do differently when given incomplete image files. I'm going to close this because it sounds like the problem is actually that Zola produces corrupt jpeg files

spl:
@fintelia:

> It isn't clear to me from this proposal what this crate could do differently when given incomplete image files.

I guess one thing that this crate could do is attempt to support bad image files to some extent, as done by various image viewers. But I infer from the comments above that this is not a goal, so most of my questions in https://github.com/image-rs/jpeg-decoder/pull/98#issuecomment-519632977 have been answered.

However, I still wonder if there's a way to repair the images such that I could use them with `jpeg-decoder`.

> I'm going to close this because it sounds like the problem is actually that Zola produces corrupt jpeg files

Just to be clear, the problem showed up for me while using Zola, which uses `jpeg-decoder`. The images are _not_ produced by Zola.

torokati44:
Have you perhaps found any fix or workaround for this yet, @spl?
I'm asking because I think https://github.com/ruffle-rs/ruffle/issues/4209 is also caused by this.

spl:
@torokati44 No, I have not. Sorry!

torokati44:
Sorry for the necrobump, but FYI, it seems that at least in our case (https://github.com/ruffle-rs/ruffle/issues/4209), checking if the data ends with an `EOI` marker, and appending it in case it doesn't, helped with the decoding: https://github.com/ruffle-rs/ruffle/pull/8820

PR #101, opened 2019-08-22
Avoid allocating a buffer and filling it just for skipping bytes

lovasoa:
This PR changes the implementation of skip_bytes : it used to allocate a buffer, fill it, and then discard it. The new implementation does not allocate.


lovasoa:
The failing test comes from crossbeam-utils, not this PR

kaksmet:
Thanks

PR #113, opened 2019-09-28
Add a method to determine the size in pixels for a given format

danieldulaney:
Sometimes it can be useful to manipulate pixels without changing their content. That code generally needs to know the size of each pixel in bytes, so that it can determine (for example) the byte offset of a line ending or the byte index of a particular pixel.

This is a utility function that makes it convenient to determine the byte width of each pixel when using a particular format.

(Note: This is essentially just the inverse of [the operation performed](https://github.com/kaksmet/jpeg-decoder/blob/c8e25252ec6c6888db8eb866dec47131c34949a3/src/decoder.rs#L95-L99) when the `PixelFormat` value is first created)

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -38,7 +38,7 @@ pub enum PixelFormat {
 
 impl PixelFormat {
     /// Determine the size in bytes of each pixel in this format
-    pub fn pixel_size(&self) -> usize {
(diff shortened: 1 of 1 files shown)


danieldulaney:
It looks like one of the dependencies (`cfg-if`) doesn't build on 1.28. It doesn't look like an issue related to this PR.

PR #115, opened 2019-10-21
parse JFIF APP0 sections and expose Density with DensityUnits

cormacrelf:
See https://github.com/image-rs/image/issues/1067

Possible improvements

- Rename to match `png::PixelDimensions`, which is not really an accurate name because it doesn't represent the size of a pixel, but the density of an image.
- Remove the thumbnail stuff completely (I added it because I was there, and maybe someone wants it? I drew the line at allocating a buffer for the packed thumbnail data.)


src/decoder.rs
```diff
@@ -58,7 +60,7 @@ pub struct Decoder<R> {
 
     restart_interval: u16,
     color_transform: Option<AdobeColorTransform>,
-    is_jfif: bool,
+    jfif_app0: Option<JfifApp0>,
```
fintelia: Are `color_transform` and `jfif_app0` mutually exclusive? If so, perhaps this could hold an AppData or Option<AppData> instead?

fintelia:
This PR generally looks good to me, though I agree that it should probably either fully decode thumbnail images or just ignore them.

typetetris:
Having the ability to get the PixelDensity from a decoded image would be really great!

What is missing here, so the MR can get merged?

fintelia:
We're no longer adding new features since the crate is in maintenance mode:

> This crate is in maintenance mode as image-rs seeks to move over to the zune-jpeg crate. However, since other Rust JPEG crates do not support lossless mode, we will continue to welcome PRs related to that functionality.

PR #116, opened 2019-10-30
fix clippy::into_iter_on_array error in decoder

scottmcm:
[Rust is trying to add `IntoIterator` for arrays](https://github.com/rust-lang/rust/pull/65819#issuecomment-547775045), which would break this.  So might as well do it the way that's forward-compatible and also shorter.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -232,7 +232,7 @@ impl<R: Read> Decoder<R> {
                 Marker::DQT => {
                     let tables = parse_dqt(&mut self.reader)?;
 
-                    for (i, &table) in tables.into_iter().enumerate() {
+                    for (i, &table) in tables.iter().enumerate() {
                         if let Some(table) = table {
                             let mut unzigzagged_table = [0u16; 64];
 


scottmcm:
CI failure doesn't look like the fault of this change:
> failed to parse manifest at `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/cfg-if-0.1.10/Cargo.toml`

CryZe:
Looks like rayon isn't actually compatible with 1.28 anymore (due to some indirect dependency on the latest cfg-if).

lovasoa:
Maybe `Cargo.lock` should be committed...

CryZe:
Nah, you don't commit Cargo.lock for libraries. Then CI would pass but it would still fail to build on 1.28 for everyone depending on this crate.

LukasKalbertodt:
@fintelia Do you think this fix can be released with a minor version bump (i.e. 0.1.17)? That would be really great, as `jpeg-decoder` caused 290 other crates to regress due to this. That's the second most regression caused after `colored`.

fintelia:
@LukasKalbertodt 0.1.17 has now been published.

PR #118, opened 2019-11-21
Bump minimum Rust version to 1.34.2

kevinmehall:
Tests have been failing for a while due to lack of support for 1.28 in rayon and its dependencies. This bumps the minimum Rust version to 1.34 to [match the `image` crate](https://github.com/image-rs/image/pull/984).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - 1.28.0
+  - 1.34.2
   - stable
   - beta
   - nightly
--- a/README.md
+++ b/README.md
@@ -32,4 +32,4 @@ fn main() {
 ```
 
 ## Requirements
-This crate compiles only with rust >= 1.28.
+This crate compiles only with rust >= 1.34.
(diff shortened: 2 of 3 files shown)


LukasKalbertodt:
@kaksmet @fintelia could you have a look at this and #116 soon? Especially the later one is a pretty straight forward fix. Lots of crates across the Rust ecosystem depend on `jpeg-decoder`, so it would be really cool if both could be merged.

And I noticed that for a few months, no PRs were merged (i.e. this crate was unmaintained for a few months). I completely understand that there are tons of reasons for why one can't put any work into open source anymore. If you read this, could you consider getting [the Rust Bus](https://users.rust-lang.org/t/bus-factor-1-for-crates/17046) on board to help out in situations like these? Thanks :)

fintelia:
@LukasKalbertodt see https://github.com/image-rs/organization/issues/9 for a bit of context. We now have [image-rs:publish](https://github.com/orgs/image-rs/teams/publish) as a crate owner, so we should hopefully be able to stay more on top of PRs going forward

PR #122, opened 2019-11-25
Add three benchmarks to the crate

lovasoa:
Test decoding of single-pass and progressive JPEGs as well as
metadata extraction.

Add "cargo bench" to the CI, and cache previous results between runs
so that performance regressions can be easily spotted.


benches/decoding_benchmark.rs
```diff
@@ -0,0 +1,35 @@
...
+
+fn read_metadata(image: &[u8]) -> ImageInfo {
+    let mut decoder = jpeg::Decoder::new(black_box(image));
+    decoder.read_info().unwrap();
+    decoder.info().unwrap()
+}
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    c.bench_function("decode a 512x512 image of colored peppers", |b| b.iter(|| {
+        read_image(include_bytes!("peppers.jpg"))
```
Bond-009: `include_bytes!` does (as the name suggests) include the bytes as a reference to a byte array. I'm afraid the compiler can do optimizations because of this that aren't possible in most real usage of this library.

lovasoa: No, the compiler doesn't do any optimization based on the contents of the byte array itself (you can look at the compiled output). 
It can't because of the `black_box` function we use.

fintelia:
Could you indicate where you got the images from / confirm that they are licensed suitably? Otherwise LGTM

lovasoa:
The peppers image is a classical image used in image compression research, it comes from the [USC-SIPI Image Database](http://sipi.usc.edu/database/). I couldn't find any licensing information attached to it. Should I replace it ?

lovasoa:
Ok, I found the information

> Sources of these are unknown and copyright status is also unknown

I'll replace the image

lovasoa:
I replaced the picture with a photograph of the Eiffel tower that I took myself, and that I am releasing free of copyright.

lovasoa:
I noticed that the benchmarks take a long time to execute on the CI. I'm going to decrease the iteration count.

lovasoa:
Done.
@fintelia , I think this is ready for merging !

PR #123, opened 2019-12-07
Bump version and changelog for v0.1.17

fintelia:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+## v0.1.17 (2019-12-07)
+- Minimum supported rust version changed to 1.34
```
HeroicKatora: A minimum support Rust bump in a minor version? I didn't realize that this already happened once a few months ago but I don't exactly know how to react to this. The previous bump might have been only due to rayon, I guess? Which feature or change necessitates this right now.

fintelia: I just went through the commits since the last release. The minimum Rust version was changed here: https://github.com/image-rs/jpeg-decoder/commit/d2cf9299c50d59f84edaae03eca7b57969eb20f5

fintelia: Per https://github.com/image-rs/jpeg-decoder/pull/116#issuecomment-562484007 we should certainly do a minor version release, but if that single line change is the only thing requiring a version bump I'd be OK with rolling it back for now

HeroicKatora: I'm pretty sure the change in that commit is not the cause, but simply dependencies being updated. Surprisingly, `rayon` specifies `1.28` [as the minimum support Rust version](https://github.com/rayon-rs/rayon/blob/99706a49b478b3cbde1135401f325556f0a4d472/Cargo.toml) so either this or the commit message is not quite accurate. The most precise I can narrow it down to is `1.30`. Trying to compile with `1.24.1` and no default features gives a few errors:

* introduction of `crate::` imports in 85adaf51d3a659cdcf6ff27c7e2ad3045cc9eb8d and 12920df36309c86368ad330444ed66c5e1e857b9 (requires 1.30)
* the `a..=b` range notation since 7feb6aa4b4db6b0d780d44cfc8d75887cc28a741 (requires 1.26)
* `dyn Trait` since 7feb6aa4b4db6b0d780d44cfc8d75887cc28a741(requires 1.27)

Also, `rayon` seems to be bumping the version requirement in SemVer compatible releases so it maybe trying to support `rayon` at the same time as a minimum Rust version is a non-starter anyways.

HeroicKatora: Since these (technically speaking) regressions have been introduced already either way, I'd do a minor version release for `1.34.1` to fix the better supported compiler version, and then try to look into fixing these few issues. That might involved having two minimum supported version policies—enfored with Travis— one for builds with rayon and one for builds without.

PR #126, opened 2019-12-09
Fix invalid line stride computation causing OOB array accesses

### LINKED ISSUE #125: Decoding broken in 0.1.17
Using image: https://i.imgur.com/nO4Jdc5.jpg (somewhat NSFW)

Code:

```rust
use std::fs::File;
use std::io::BufReader;

fn main() {
    let file = File::open("nO4Jdc5.jpg").expect("failed to open file");
    let mut decoder = jpeg_decoder::Decoder::new(BufReader::new(file));
    let pixels = decoder.decode().expect("failed to decode image");
    let metadata = decoder.info().unwrap();
}
```

Works with version `=0.1.16`, crashes with version `=0.1.17`. Error is `thread 'main' panicked at 'index out of bounds: the len is 1010816 but the index is 1015928'`. Seems to be a bug in single-channel decoding.

lovasoa:
Fixes #125


kevinmehall:
Thanks for fixing my mistake!

There should probably be a non-square test case to catch width/height issues like the second commit, and it might as well be a reftest since it's a valid image.

fintelia:
This looks good to me. Could you add a changelog entry + version bump to Cargo.toml, so we can push out a fix right away?

lovasoa:
@fintelia : done.

PR #127, opened 2019-12-09
Optimize data copy

lovasoa:
Data was copied byte by byte in two nested loops.
This changes the code to a single loop, using memcpy to copy one line of data at a time.

I observed a 2% performance improvement on the "decode 512x512 image" benchmark on a 2018 Macbook Pro.

This was branched from #126.


src/decoder.rs
```diff
@@ -785,12 +785,14 @@ fn compute_image(components: &[Component],
...
-        for y in 0 .. width {
-            for x in 0 .. height {
-                buffer[y * width + x] = data[0][y * line_stride + x];
-            }
+        let line_stride = component.block_size.width as usize * component.dct_scale;
+
+        for y in 0 .. height {
+            let destination_idx = y * width;
+            let source_idx = y * line_stride;
+            let destination = &mut buffer[destination_idx..destination_idx + width];
```
fintelia: One trick you can do here to avoid repeating destination_idx is to use `buffer[destination_idx..][..width]`

lovasoa:
@fintelia : Ok, I updated indexing.

PR #128, opened 2019-12-10
chore: Update png to 0.15

ignatenkobrain:


HeroicKatora:
Was done as part of the larger #136, thanks in any case!

PR #131, opened 2020-02-12
Add afl fuzz target

dbrgn:
Add a decoder fuzz target using afl.rs.

The fuzzer already found panics within seconds of running it :slightly_smiling_face: I'll open separate issues for those.


Shnatsel:
Awesome! You should try using the sets of images generated by AFL on other libraries as input to speed things up: http://lcamtuf.coredump.cx/afl/demo/

Also, you might want to add the bugs you've found to https://github.com/rust-fuzz/trophy-case. This not only gets you perpetual glory, but also enables interesting quantitative analysis like [this one](https://www.reddit.com/r/rust/comments/f1ynel/quantitative_data_on_the_safety_of_rust/).

dbrgn:
@Shnatsel thanks, I wasn't aware of those input corpora! Will update the PR.

> Also, you might want to add the bugs you've found to https://github.com/rust-fuzz/trophy-case

Yep, I will submit all confirmed bugs.

dbrgn:
Added 98 new inputs from http://lcamtuf.coredump.cx/afl/demo/, as well as a second fuzz target that parses the header only (`read_info`). Decoding will parse the header as well, but it might still be interesting to fuzz the header parsing separately.

fuzz-afl/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+    $ mkdir out/
+    $ cargo afl fuzz -i in/ -o out/ target/release/fuzz_<target>
+
+To reproduce a crash:
+
+    $ cargo run --bin reproduce_<target>
+
+Note: You should also try fuzzing in debug mode, since things like overflow
+checks don't happen in release mode. (Release mode is much faster though.)
```
Shnatsel, line 23: `cargo afl build` actually compiles release mode with debug assertions and overflow checks.

dbrgn:
@HeroicKatora oh, sorry for forgetting about this PR. thanks for the update!

HeroicKatora:
> Sorry for forgetting about this PR. thanks for the update!

It's basically my responsibility to to reminders once in a while. Consider yourself reminded :wink: 

Thank you!

PR #100, opened 2019-08-09
Extraneous bytes after SOS

filipl:
I have a scanner (brother DSmobile 920DW) that is very generous with the bytes it outputs. There is as far as I can see almost extra bytes and RST markers after the expected last one.

If I convert with imagemagick (using libjpeg-turbo) I get the following output:
```
convert: Corrupt JPEG data: 2634 extraneous bytes before marker 0xd2 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 616 extraneous bytes before marker 0xd3 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 650 extraneous bytes before marker 0xd4 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 599 extraneous bytes before marker 0xd5 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 585 extraneous bytes before marker 0xd6 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 2634 extraneous bytes before marker 0xd7 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 616 extraneous bytes before marker 0xd0 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 650 extraneous bytes before marker 0xd1 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
convert: Corrupt JPEG data: 599 extraneous bytes before marker 0xd9 `extraneous-bytes-after-sos.jpg' @ warning/jpeg.c/JPEGWarningHandler/398.
```
With my changes (and some additional printouts that I didn't include in the commit since I'm not a fan of libraries writing to stdout/stderr) I can get jpeg-decoder to output the same messages. The changes should make the finding of the next marker fairly similar to how libjpeg and libjpeg-turbo does it.

Looking in the JPEG specification it seems as if these changes should be the correct approach.
> B.1.1.2
[...]

PR #147, opened 2020-04-22
Revert 0xFF marker check

ggopcceop:
I has some webcam from different venders that i use v4l2 to capture mjpg images to do some processing. When i used image-rs to decode some of the images, i got `Error: IoError(Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" })`. 

After some investigation i found that the 
https://github.com/image-rs/jpeg-decoder/blob/0c6f5a20801c9530fb99487a69ba39037ce15d0c/src/decoder.rs#L352
 function skipped EOI marker when there is multiple 0xff before 0xD9. 

And there is pull request from #100 that removed code to skip optional 0xff. So I added it back.

There is also a test image i took that have this error before. but I'm not sure where to put this image in the tests dir, since the crash test doest care decode result.


ggopcceop:
Sorry about the confusion. And yes, I took that image my self, you have my consent to use the image freely.

PR #150, opened 2020-04-25
Optimize final grayscale image creation

lovasoa:
This is based on #149, so please merge this PR  only after #149.

This pr introduces two small changes that avoid unneeded computation and data copying when decoding grayscale images.


HeroicKatora:
And now you'll need to rebase this anyways due to the squash :/

lovasoa:
I rebased and added another commit to reuse the decoded image data instead of allocating a new buffer. The performance would be better if we could use copy_within from rust 1.37.

HeroicKatora:
This is not always a performance win. The timings on my machine:

```
[Benchmark]                      	[Change] 	[Base]  	[Head]  	[Significance]
decode a 512x512 JPEG            	improved 	3.434027	3.217900	0.050000
decode a 512x512 progressive JPEG	regressed	5.677390	6.155000	0.050000
decode a 512x512 grayscale JPEG  	improved 	950.903415	914.770000	0.050000
extract metadata from an image   	         	985.927390	998.270000	0.050000
```

HeroicKatora:
I've rebase the PR to have
1. one commit introducing the new benchmark
2. two consecutive commits with the performance changes
3. and the bug fix

This should make it easier for others to try out the benchmarks.

lovasoa:
Yes, the initial idea was to use copy_within, which would have avoided the double data copy, but then I realized that this library must work with old compiler versions.

Also, about the benchmarks : there is a performance difference between images with a width that is a multiple of the block size (like the one I added to the benchmarks) and the others.

lovasoa:
We may want to try [rotate_left](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.rotate_left)... Do you want me to make a new commit with that ?

HeroicKatora:
> We may want to try rotate_left... Do you want me to make a new commit with that ?

Would be worth a try, yes.

lovasoa:
I tried the `rotate_left` approach : 

```rust
        if usize::from(output_size.width) != line_stride {
            // The first line already starts at index 0, so we need to move only lines 1..height
            for y in 1..height {
                let destination_idx = y * width;
                let source_idx = y * line_stride;
                // We could use copy_within, but we need to support old rust versions
                let diff = source_idx - destination_idx;
                decoded[destination_idx..source_idx+width].rotate_left(diff);
            }
        }
```

and benchmarked it by converting tower_greyscale.jpg to 1607x1607. The version with rotate_left is ~3% slower. So I didn't add it to this PR. I think this is ready to merge.

HeroicKatora:
What about the regression of the progressive decoding? Did it not regress on your machine?

lovasoa:
Yes, the progressive decoding hasn't changed at all.

HeroicKatora:
I must have measured during high noise on my system (maybe background update?). Re-did them and as expected only the grayscale decoding improved and no regressions :tada:

PR #156, opened 2020-06-24
Reuse allocations in upsampling for a slight performance boost

Shnatsel:
Cuts the total number of allocations in half.

This matters surprisingly little. On glibc I get no difference and on musl just 3% improvement. I'm not sure if this is worth the trouble.


Shnatsel:
Closing since this increases complexity and doesn't result in any measurable performance improvement.

PR #159, opened 2020-06-27
refactor: Extract and clarify the 8x8 idct implementation

Marwes:
This code is extracted from https://github.com/Marwes/combine-jpeg which
is originally based on this repo. While the parser parts are mostly
rewritten, the parts like idct, huffman etc are mostly the same, though
with some performance and (in my opinion) clarity improvements to the
code.

If this PR is accepted I may try and extract some other parts as well.

(While this PR doesn't contain any performance improvements, I will just
add that `combine-jpeg` has some significant performance improvements.
It still lacks some features but it's almost, but not quite as fast as
mozjpeg on the images that it accepts (no simd for either). Some of this
is because of unsafe, some of it because of less overhead in the
decoding).

```
it_works/combine        time:   [2.6562 ms 2.6633 ms 2.6749 ms]
                        change: [-7.2890% -1.9145% +1.5656%] (p = 0.68 > 0.05)
                        No change in performance detected.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high severe
it_works/mozjpeg        time:   [2.0400 ms 2.0415 ms 2.0429 ms]
                        change: [+0.4243% +0.6680% +0.8305%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
it_works/jpeg-decoder   time:   [8.9513 ms 9.1526 ms 9.3654 ms]
                        change: [-13.663% -5.0961% +1.6086%] (p = 0.29 > 0.05)
                        No change in performance detected.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low severe
  1 (10.00%) high mild

```

### DIFF
--- a/src/idct.rs
+++ b/src/idct.rs
@@ -2,7 +2,10 @@
 // One example is tests/crashtest/images/imagetestsuite/b0b8914cc5f7a6eff409f16d8cc236c5.jpg
 // That's why wrapping operators are needed.
 use crate::parser::Dimensions;
-use std::num::Wrapping;
+use std::{
+    convert::TryFrom,
+    num::Wrapping,
+};
 
 pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions) -> usize {
     fn scaled(len: u16, scale: usize) -> u16 { ((len as u32 * scale as u32 - 1) / 8 + 1) as u16 }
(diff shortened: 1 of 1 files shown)


src/idct.rs
```diff
@@ -130,72 +128,126 @@ fn dequantize_and_idct_block_8x8(coefficients: &[i16], quantization_table: &[u16
...
-        output[i * output_linestride + 7] = stbi_clamp((x0 - t3) >> 17);
-        output[i * output_linestride + 1] = stbi_clamp((x1 + t2) >> 17);
-        output[i * output_linestride + 6] = stbi_clamp((x1 - t2) >> 17);
-        output[i * output_linestride + 2] = stbi_clamp((x2 + t1) >> 17);
-        output[i * output_linestride + 5] = stbi_clamp((x2 - t1) >> 17);
-        output[i * output_linestride + 3] = stbi_clamp((x3 + t0) >> 17);
-        output[i * output_linestride + 4] = stbi_clamp((x3 - t0) >> 17);
+        const X_SCALE: i32 = 65536 + (128 << 17);
+
+        let [s0, rest @ ..] = chunk;
```
HeroicKatora, line 142: For compatibility with `1.34.2` we need to adjust this pattern slightly to use the method instead. It shouldn't make a difference in the final code as `chunk` is an array type with statically known size.

```suggestion
        let (s0, rest) = chunk.split_first().unwrap();
```

HeroicKatora:
Thanks for contributing back some of your changes :)

PR #160, opened 2020-07-01
Decode exactly completed components of a scan

HeroicKatora:
Fixes a bug where components were sometimes missing, when the completion
was spread between different scans. It does not affect the decoding of
images when all components are completed in the same scan.

This is not a complete fix (as noted by the still failing sample) but should correctly 
decode many more images. In particular, we can now decode some progressive
jpegs in which the final AC data is split among multiple scans. What we still can
not do correctly is decode images where some components specify only a subset
of all 63 coefficients. In the words of the spec, we can do most of spectral selection.
Additionally, successive approximation is also partially not yet supported.


HeroicKatora:
This fixes, with some confidence, the structural issues of the decoding. Some color issues might still be present if encoders make full use of the allowed band splitting of ACs and not only the DC/AC split tested here and apparently present in the samples provided. This would require a larger rework of the complete decoding flow, so let's fix the error and very, very common case first.

PR #163, opened 2020-10-03
Propagate "invalid dimensions" error

### LINKED ISSUE #161: Panic with "invalid dimensions" in decode
When trying to decode the image, it fails with

```
thread 'main' panicked at 'invalid dimensions', src/parser.rs:245:5
[...]
  11: std::panicking::begin_panic
             at /rustc/4fb7144ed159f94491249e86d5bbd033b5d60550/src/libstd/panicking.rs:397
  12: jpeg_decoder::parser::ceil_div
             at src/parser.rs:245
  13: jpeg_decoder::parser::update_component_sizes
             at src/parser.rs:255
  14: jpeg_decoder::parser::parse_sof
             at ./src/parser.rs:228
```

invalid-dimensions

hanmertens:
Avoid panicking when decoding an image with invalid dimensions but propagate the error instead. This required changing some function signatures, but none of these functions are part of the public API. I kept the error message the same, let me know if it should be changed to something more descriptive.

Fixes #161


src/parser.rs
```diff
@@ -241,29 +243,31 @@ pub fn parse_sof<R: Read>(reader: &mut R, marker: Marker) -> Result<FrameInfo> {
 }
 
 /// Returns ceil(x/y), requires x>0
-fn ceil_div(x: u32, y: u32) -> u16 {
-    assert!(x>0 && y>0, "invalid dimensions");
-    (1 + ((x - 1) / y)) as u16
+fn ceil_div(x: u32, y: u32) -> Result<u16> {
+    if x ==0 || y == 0 {
+        return Err(Error::Format("invalid dimensions".to_owned()));
```
HeroicKatora, line 248: When I opened the issue report, I was unsure how the code even reached this point in the first place. At least I was under the impression that both numbers were checked individually before. It doesn't really matter. However, would you add a `TODO` to validate all inputs more closely such that we might error earlier?

HeroicKatora (commented):
Please add the file as a regression test case. Simply drop it into `tests/crashtest`. It's synthetic by the fuzzer (ran by me) so no copyright issues :)

HeroicKatora:
Looks good :+1:

src/parser.rs
hanmertens: Sure, will do!

HeroicKatora (approved):

HeroicKatora:
Thank you!

PR #167, opened 2020-10-17
Eliminate some bounds checks in IDCT

Shnatsel:
Convert a debug assert into an actual assert to use as an optimizer hint. Eliminates a bunch of bounds checks.

It reduces the number of assembly instructions, but I did not see any measurable performance improvement on my machine.

Instruction counts for `jpeg_decoder::worker::immediate::ImmediateWorker::append_row_immediate` where this gets inlined: before - 1420, after - 1363.

Top 10 instructions with counts, before:

```
    385 mov
    150 lea
     95 add
     86 imul
     85 cmp
     61 jmp
     50 sar
     49 sub
     36 call
     33 movzx
```

Instruction counts after:

```
    355 mov
    150 lea
     91 add
     86 imul
     83 cmp
     53 jmp
     50 sar
     48 sub
     35 call
     34 movzx
```

### DIFF
--- a/src/idct.rs
+++ b/src/idct.rs
@@ -2,10 +2,7 @@
 // One example is tests/crashtest/images/imagetestsuite/b0b8914cc5f7a6eff409f16d8cc236c5.jpg
 // That's why wrapping operators are needed.
 use crate::parser::Dimensions;
-use std::{
-    convert::TryFrom,
-    num::Wrapping,
-};
+use std::{convert::TryFrom, convert::TryInto, num::Wrapping};
 
 pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions) -> usize {
     fn scaled(len: u16, scale: usize) -> u16 { ((len as u32 * scale as u32 - 1) / 8 + 1) as u16 }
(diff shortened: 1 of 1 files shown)


Shnatsel:
In parallel mode this speeds up decoding from 320ms to 300ms, or at about 1%. No change when rayon is not enabled because the IDCT worker thread is not in the critical path then.

HeroicKatora:
I wonder how complex it would be to funnel through the precise type `&[i16; 64]`?

HeroicKatora:
Nevermind, that obviously requires `bytemuck` or probably const generics..

HeroicKatora (approved):

Shnatsel:
A lot of precise types can and should be funnelled through in theory. That match on a `usize` is gross, instead of panicking here it should just be an enum with custom discriminants.

HeroicKatora:
> A lot of precise types can and should be funnelled through in theory. That match on a usize is gross, instead of panicking here it should just be an enum with custom discriminants.

There are probably a few other types that are not properly newtype wrapped or should be enums, yeah.

PR #168, opened 2020-10-22
Multi-threaded IDCT

Shnatsel:
Spawn a thread for each color component in IDCT. Should in theory speed up IDCT by 3x and the entire process by like 2x, but only speeds up end-to-end decoding a bit because we get bottlenecked by the reader thread instead.

The code probably isn't the cleanest, since I wrote this at 5AM, so I'm open to feedback here. The code I started with also wasn't the cleanest, but that's not much of an excuse.

I've run `cargo test` but haven't tested this on real-world images.


Shnatsel:
I've tested this on a few images I had lying around, and it seems to be good to go.

The only concern I have is that we parallelize up to 4 threads unconditionally up from the previous default of 2 threads, but seeing that threads are very cheap (I can spawn and joun 50,000 of them per second on a single core) and there's plenty of memory allocation happening regardless for a typical image, I don't see that being a problem.

HeroicKatora (approved):

Issue #173, opened 2020-11-23
Some images get corrupted during decoding


mwap:
Some images like this one:
https://ant-photo.eu/wp-content/uploads/2020/07/Cataglyphis-iberica-worker2.jpg
end up looking like this:
https://i.ibb.co/tp2d6Wy/out.png
after simply decoding the image and saving the output in a human readable format.

Example program:
```toml
[dependencies]
bmp = "0.5.0"
jpeg-decoder = "0.1"
```
```rust
extern crate jpeg_decoder as jpeg;
#[macro_use]
extern crate bmp;

use std::fs::File;
use std::io;
use bmp::*;

fn main() {
    let file = File::open("broken.jpg").unwrap();
    let mut dec = jpeg::Decoder::new(io::BufReader::new(file));
    let pixels = dec.decode().unwrap();
    let mut img = Image::new(1920, 1080);
    for s in 0..1920 * 1080 {
        img.set_pixel(s % 1920, s / 1920, px!(pixels[s as usize * 3], pixels[s as usize * 3 + 1], pixels[s as usize * 3 + 2]));
    }
    img.save("out.bmp");
}
```

This happens to some of the other images at https://ant-photo.eu/galeria/zdjecia-mrowek

mwap:
Worth noting that every other software I have used to process these images with has done so without any problems

kaj:
I get the same (broken) results with the `decode` example in this crate.  The original image looks good in eog (gnome image viewer) and firefox.  So yes, this seems to be a bug in jpeg-decoder.

According to the linux `file` utility, the image trigging the bug is:
> broken.jpg: JPEG image data, JFIF standard 1.01, resolution (DPI), density 72x72, segment length 16, progressive, precision 8, 1920x1080, components 3

quilan1:
It looks like the error is first occurring on the second segment of the image (Y component, spectral components 1..6).

The data in the worker.jpg file there (offset 0x26E6) is: `FE E1 FF D0 FE E1`. The first two bytes corresponds to the huffman encoding for 0x70 (EOB for the next 7-bits' worth of MCUs, with an invisible tacked on '1' at the beginning) or: `11111110` (huffman code for 0x70) `[1]1110000` (240 MCUs long) `1` (unused bit). So, the next 240 MCUs should be skipped. This is also the reset (RST) length, which corresponds to the row size in blocks (1920 / 8).

[...]

Issue #89, opened 2018-07-03
Potential bug decoding non-interleaved scan with RST enabled.


liuzjmike:
The standard says that when a thread is non-interleaved the MCU is just on block. But decode_scan still treats the MCU as H*V blocks when looking for RST markers. Is this a bug?

PR #176, opened 2021-01-14
Bugfix non interleaved mcus

quilan1:
As detailed in the commits, the issue had to do with the way the old code treated MCUs in non-interleaved streams, with channels having a greater-than-1 sampling factor. In a non-interleaved stream, MCUs should be treated as single blocks instead of as the group of HxV blocks.

Future work might be considered to simplify the coding of batching rows together for the worker threads; currently it only accepts the full HxV set of blocks for a 'row', instead of an image row, and I didn't touch that code.

Solves #173 and addresses #89


quilan1:
It occurs to me that the created jpeg file causes an error (bad huffman decode), not a panic; would this be better placed in the reftest images?

HeroicKatora:
Yes, `crashtest` is for regressions that previously panicked and ignores the result. If you want to test success or a particular decoding response then the reftests are the appropriate place.

HeroicKatora (approved):
Mental note: At some future point we should change the `as u16` casts to `u16::from` where applicable because it makes it clear that they are lossless. I did stumble on this in the review.

HeroicKatora:
Thank you!

Issue #42, opened 2016-07-05
question: exif support?


liamstask:
I've been doing some testing with this library, which seems great, and have a couple questions:
* I've made a couple hacks locally to identify exif data - would you be interested in a PR adding this functionality and, if so, how would you prefer this to be structured? My default was to simply add a `Vec<u8>` to the decoder, and allow users to process it outside the library as they please.
* ultimately I would like to edit some of the exif and write the jpeg back out with the updated exif. i haven't been able to immediately find any projects that support encoding a jpeg with exif data - don't suppose you have any recommendations? :D

kaksmet:
It would be great to receive a PR for getting exif data. Exposing a `Vec<u8>` sounds good to me, I have plans to do the same with embedded icc profiles.

As for your other question, do you only want to change the exif data segment or do you also need to reencode the image?

liamstask:
thanks for the quick response. and great - i will try to put a PR together and submit it.

for the 2nd question, yes I'd only want to modify the exif segment, and then write it back out along with the untouched image data.

kaksmet:
For just modifying the exif data, I would split the file at the beginning and end of the exif data segment and then just swap in the modified exif data.

However if you only change the values in the exif data and don't add/remove anything you could just modify it in-place.

liamstask:
Hm, the issues I see in that case are:
* the decoder interface would need to return the offsets for the exif data, in addition to the `Vec<u8>` of the data itself - maybe this is OK?
* in order to write the exif segment back out, my application code would need to know the format/layout of a jpeg segment. I suppose application code can do that, but would be better handled by a jpeg encoder lib, if possible.

kaksmet:
>  the decoder interface would need to return the offsets for the exif data, in addition to the Vec<u8> of the data itself - maybe this is OK?

The easiest way to find the offset of a JPEG segment is to use something like [immeta's `find_marker`](https://github.com/netvl/immeta/blob/87db7b9372a9ec997753ecb714d6b6c71ab79eed/src/formats/jpeg.rs#L92-L106) function, searching for a APP1 marker (which has a value of `0xE1`). Right at the start of the segment is it's length as a big-endian u16 (which includes the size of itself, so you'll want to subtract 2 from it).

> in order to write the exif segment back out, my application code would need to know the format/layout of a jpeg segment. I suppose application code can do that, but would be better handled by a jpeg encoder lib, if possible.

[...]

PR #182, opened 2021-01-31
Add support for extracting Exif data

fseegraeber:
Implements #42.

Also a first step towards [image#1045](https://github.com/image-rs/image/issues/1045).

Exposes Exif data as an `Option<Vec<u8>>` that can be parsed by client code e.g. using `kamadak-exif` (or simply to pass it on when saving the file again):

```rust
    let path = Path::new("tests")
        .join("reftest")
        .join("images")
        .join("ycck.jpg");

    let mut decoder = jpeg::Decoder::new(File::open(&path).unwrap());
    decoder.decode().unwrap();

    let exif_data = decoder.exif_data().unwrap();
    let exif_reader = exif::Reader::new();
    let exif = exif_reader.read_raw(exif_data).unwrap();
    for field in exif.fields() {
        println!("{}", field.display_value());
    }
```

Not 100% sure about the `exif_data()` method. Since Exif might contain up to 64kb a copy might be a bit much.
Perhaps:
```rust
pub fn take_exif_data(&mut self) -> Option<Vec<u8>> {
    self.exif_data.take()
}
```
or
```rust
pub fn exif_data(&self) -> Option<&Vec<u8>> {
    self.exif_data.as_ref()
}
```

I guess exif data might be inspected multiple times e.g. to determine orientation?

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -75,6 +75,8 @@ pub struct Decoder<R> {
 
     icc_markers: Vec<IccChunk>,
 
+    exif_data: Option<Vec<u8>>,
+
     // Used for progressive JPEGs.
     coefficients: Vec<Vec<i16>>,
     // Bitmask of which coefficients has been completely decoded.
@@ -95,6 +97,7 @@ impl<R: Read> Decoder<R> {
             is_jfif: false,
             is_mjpeg: false,
             icc_markers: Vec::new(),
+            exif_data: None,
             coefficients: Vec::new(),
             coefficients_finished: [0; MAX_COMPONENTS],
         }
@@ -122,6 +125,13 @@ impl<R: Read> Decoder<R> {
             },
             None => None,
         }
+    }
+
+    /// Returns raw exif data, starting at the TIFF header, if the image contains any.
+    ///
+    /// The returned value will be `None` until a call to `decode` has returned `Ok`.    
+    pub fn exif_data(&self) -> Option<Vec<u8>> {
+        self.exif_data.as_ref().map(|v| v.clone())
     }
 
     /// Returns the embeded icc profile if the image contains one.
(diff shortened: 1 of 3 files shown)


src/decoder.rs
```diff
@@ -124,6 +127,13 @@ impl<R: Read> Decoder<R> {
         }
     }
 
+    /// Returns raw exif data, starting at the TIFF header, if the image contains any.
+    ///
+    /// The returned value will be `None` until a call to `decode` has returned `Ok`.    
+    pub fn exif_data(&self) -> Option<Vec<u8>> {
+        self.exif_data.as_ref().map(|v| v.clone())
```
SirWindfield, line 134: Wouldn't `Option<&[u8]>` make more sense? `self.exif_data.as_ref()` only. That way you don't need to copy the content the whole time. If people need an owned type, they can simply convert the slice to a vec themselves 🤔

fseegraeber: You mean `as_deref()` right? So option 3:
```rust
pub fn exif_data(&self) -> Option<&[u8]> {
        self.exif_data.as_deref()
    }
```
Sorry for the rookie mistakes. 😄

fseegraeber: Nevermind, different rust version.

HeroicKatora: A `match` works as well:
```rust
match &self.exif_data {
    None => None,
    Some(data) => Some(data.as_slice()),
}
```

fseegraeber: I initially went with 
```rust 
self.exif_data.as_ref().map(|v| v.as_ref())
```
but `as_slice()` seems a bit more obvious.

SirWindfield: Is there a MSRV? Because if not, that could be written using `as_deref`, which is way cleaner

fseegraeber: There are CI checks that run on 1.34.2, they fail when using `as_deref`. Not sure if there is a policy though.

fintelia: I believe that 1.34.2 is our MSRV. We occasionally bump that, but I don't think it is worth it in this case

HeroicKatora:
I must have missed the force pushed commit addressing the review comment. @fseegraeber Pinging a maintainer when a PR is in limbo is absolutely encouraged :slightly_smiling_face:

fseegraeber:
Will do! 😄

PR #183, opened 2021-01-31
build: pin rust version

SirWindfield:
This makes development a little bit easier as it forces everybody 
to work with the MSRV instead of `stable` locally.


SirWindfield:
It's still possible to overwrite the toolchain version by invoking commands using Rust's buildin overwrite mechanism: `cargo +stable ...`.

HeroicKatora:
I didn't realize that this exists, until now :grinning:

PR #193, opened 2021-07-31
Lossless JPEG decoding

vsaase:
Hi,

this adds support for lossless decoding. As this is a special need mostly in medical imaging I tried to touch as little code as possible.
Performance is a bit low at the moment, help would be appreciated here.

If you think this is too special a use case I might also publish it in its own crate.


src/decoder/lossless.rs
```diff
@@ -0,0 +1,212 @@
...
+};
+use std::io::Read;
+
+impl<R: Read> Decoder<R> {
+    /// decode_scan_lossless
+    pub fn decode_scan_lossless(
+        &mut self,
+        frame: &FrameInfo,
+        scan: &ScanInfo,
+    ) -> Result<(Option<Marker>, Option<Vec<Vec<u16>>>)> {
```
HeroicKatora, line 19: I can't see any return path making use of `Option<Vec<_>>` here. In any case, when the types become this complicated I found it better to create a very simple struct and be it just for the ability to properly document each portion of the return type as field documentation.

src/decoder/lossless.rs
```diff
@@ -0,0 +1,212 @@
...
+                        scan.point_transform,
+                        frame.precision,
+                        mcu_x,
+                        mcu_y,
+                        self.restart_interval > 0
+                            && mcus_left_until_restart == self.restart_interval - 1,
+                    );
+                    let result = ((prediction + diff) & 0xFFFF) as u16; // modulo 2^16
+                    results[i].push(result << scan.point_transform);
+                }
```
HeroicKatora, line 126: This is probably the cause of the slowness because there are countless checked indices here and none of these computations can be effectively transformed to SIMD loops. We can work on this later though but a comment would make it clear we're aware of it.

src/decoder/lossless.rs
```diff
@@ -0,0 +1,212 @@
...
+                        scan.predictor_selection,
+                        scan.point_transform,
+                        frame.precision,
+                        mcu_x,
+                        mcu_y,
+                        self.restart_interval > 0
+                            && mcus_left_until_restart == self.restart_interval - 1,
+                    );
+                    let result = ((prediction + diff) & 0xFFFF) as u16; // modulo 2^16
+                    results[i].push(result << scan.point_transform);
```
HeroicKatora, line 125: This push happens unconditionally, so we are aware of the require memory for each of the result vectors. In other words, they can be pre-allocated with `Vec::with_capacity` before.

src/decoder/lossless.rs
```diff
@@ -0,0 +1,212 @@
...
+    } else {
+        let mut decoded: Vec<u16> = vec![
+            0u16;
+            ncomp * output_size.width as usize * output_size.height as usize
+        ];
+        for (i, component) in components.iter().enumerate() {
+            for x in 0 .. data[i].len() {
+                decoded[x * ncomp + i] = data[i][x];
+            }
+        }
```
HeroicKatora, line 198: Also slower than necessary. We're passing over `decoded` slice a total of `ncomp` times, each time writing single bytes into larger chunks. This should be the other way around:

```suggestion
        for (x, chunk) in decoded.chunks_mut(ncomp).enumerate() {
            for (i, (component_data, _)) in data.iter().zip(&components).enumerate() {
                chunk[i] = component_data[x];
            }
        }
```

src/decoder/lossless.rs
```diff
@@ -0,0 +1,212 @@
...
+impl<R: Read> Decoder<R> {
+    /// decode_scan_lossless
+    pub fn decode_scan_lossless(
+        &mut self,
+        frame: &FrameInfo,
+        scan: &ScanInfo,
+    ) -> Result<(Option<Marker>, Option<Vec<Vec<u16>>>)> {
+        let ncomp = scan.component_indices.len();
+        assert!(ncomp <= MAX_COMPONENTS);
+        let mut results = vec![Vec::new(); MAX_COMPONENTS];
```
HeroicKatora, line 22: Should this not have a length of `scan.component_indices.len()`?

tests/reftest/images/lossless/LICENSE
```diff
@@ -0,0 +1,20 @@
...
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
```
HeroicKatora, line 20: Just to be clear, I'm somewhat careful of MIT licensed JS packages because I believe it is some tool's default license and general license awareness isn't too great (I've seen several examples of improper licenses already). It's also not clear if the code license is supposed to apply to it's reference images in the first place. That said, we're not redistributing those images in the packaged `crate` so that's probably good enough as is.

tests/reftest/mod.rs
```diff
@@ -49,36 +49,55 @@ fn reftest_decoder<T: std::io::Read>(mut decoder: jpeg::Decoder<T>, path: &Path,
...
     ref_reader.next_frame(&mut ref_data).expect("png decode failed");
     let mut ref_pixel_format = ref_info.color_type;
 
-    if ref_pixel_format == png::ColorType::RGBA {
+    if ref_pixel_format == png::ColorType::RGBA { 
         ref_data = rgba_to_rgb(&ref_data);
         ref_pixel_format = png::ColorType::RGB;
     }
+    let mut refdata_16 : Vec<u16> = ref_data.iter().map(|x| *x as u16).collect();
+    let mut data_u16 : Vec<u16> = data.iter().map(|x| *x as u16).collect();
```
HeroicKatora, line 73: If you move the initial assignment to the two branches `L8` and `RGB24`—accepting that this will duplicate it—Rust will happily allow you to remove the initial assignment and `mut` qualifier here.

tests/reftest/mod.rs
```diff
@@ -49,36 +49,55 @@ fn reftest_decoder<T: std::io::Read>(mut decoder: jpeg::Decoder<T>, path: &Path,
...
-        jpeg::PixelFormat::RGB24 => assert_eq!(ref_pixel_format, png::ColorType::RGB),
+        jpeg::PixelFormat::L8 => {
+            assert_eq!(ref_pixel_format, png::ColorType::Grayscale);
+            assert_eq!(ref_info.bit_depth, png::BitDepth::Eight);
+        },
+        jpeg::PixelFormat::L16 => {
+            assert_eq!(ref_pixel_format, png::ColorType::Grayscale);
+            assert_eq!(ref_info.bit_depth, png::BitDepth::Sixteen);
+            refdata_16 = ref_data.chunks_exact(2).into_iter().map(|a| u16::from_be_bytes([a[0],a[1]])).collect();
+            data_u16 = data.chunks_exact(2).into_iter().map(|a| u16::from_be_bytes([a[0],a[1]])).collect();
```
HeroicKatora, line 84: `into_iter()` is superfluous here.

HeroicKatora (commented):
I'm pretty happy with this in total, except that the numerical part of prediction isn't very SIMD friendly (which is probably the reason for the slowness). This is necessarily a breaking change due to the additional pixel formats that are being introduced which makes it really easy to justify the MSRV bump. Just one question, can you specify precisely where the `1.40` requirement comes from?

examples/decode.rs
```diff
@@ -1,5 +1,9 @@
 extern crate jpeg_decoder as jpeg;
 extern crate png;
+extern crate core;
+
+use core::mem::size_of;
+use core::mem::forget;
```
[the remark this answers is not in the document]
vsaase: yes, removed it

src/decoder/lossless.rs
vsaase: It's the same return type the `decode_scan` function in `decoder.rs` has. However I now removed the Option around `Vec<Vec<u16>>`.

src/decoder/lossless.rs
vsaase: thanks, I added a comment. There is probably some redundancy in checking the indices and the selection of the predictor. However I think this prediction method is generally not well suited to parallelization, since almost all computations depend on the immediate previous results.

src/decoder/lossless.rs
vsaase: done

src/decoder/lossless.rs
vsaase: thanks, that works

src/decoder/lossless.rs
vsaase: right

tests/reftest/mod.rs
HeroicKatora: Not sure if have realized Rust can do this but you can declare the variables without initialization. As long as you initialize them in each branch of the match the compiler will happily treat them as initialized afterwards.

```rust
let refdata_16: Vec<u16>;
let data_u16: Vec<u16>;

match pixel_format {
    L8 => {
        refdata_16 = …;
        data_u16 = …;
    }
    …
}
```

No need to change the code another time but sometimes this is clearer.

src/decoder/lossless.rs
HeroicKatora: Gotcha, I see why it was initially chosen. Although, this increases the benefit of having a new type defined for this instance since it could be shared between methods.

src/decoder/lossless.rs
HeroicKatora: While it is true that there is a data dependency between result values of a single coding unit, the exact concurrency is between _different_ coding units. I believe the only dependency is within a coding unit. In other words, distinct coding units can be perfectly parallelized. Also note that by outlining the choice of prediction algorithm we free up to 2 registers for the simple predictor cases which could double the effective throughput by doing twice as many coding units at once.

HeroicKatora: I'm pretty confident that we can exploit instruction level parallelism if the prediction loop was on its own, and the selection of algorithm happened before. Note that for some of the predictor paths we effectively ignore 2-3 of the values passed in. In other words, those are useless loads _and_ they occupy register space.

Additionally, this interleaving of decoding and computation does not mix too well with bounds checks since it adds extra complication for the compiler to notice that all accesses are in bounds. If the length of `result[i]` was a constant by that point this would be easier. [See here for a similar change that had up 30% throughput gains](https://github.com/image-rs/image-png/pull/198/files#diff-94c4ec4f5acf5bd713712003e0b0ece62186fa93a897a1a42f977496fc120514).

There are some predictors that parallelize well, such as `Rb` and `Rc` which only depends on the previous line that is finished completely beforehand. Lastly, a data dependency doesn't rule out the use of simd entirely although by this point it would probably need to be done by hand instead of relying on auto-vectorization. In images with large enough lines (greater than SIMD width) only the input `ra` changes. We can calculate the prediction of a chunk of pixels at once using an assumed input of `0` and then iterate pixels and, once we know the true input, update the prediction. This still makes some of the computation parallel. For example in `RaRbRc?` predictors the update is simply delaying the add of `ra`. This _should_ provide additional throughput in total.

vsaase: Thanks for the ideas. The by far most common predictor seems to be `Ra` (default output of DCMTK library JPEG encoder). On the first column it is always `Rb`. Thus we could first compute the first column and then in parallel compute all the rows.

vsaase:
> Just one question, can you specify precisely where the `1.40` requirement comes from?

This is not related to my changes. When running `cargo check` in the master branch with the current version `1.34.2` under Windows I get this error in crossbeam-utils, which is a dependency of rayon:
`error[E0161]: cannot move a value of type dyn std::ops::FnOnce() + std::marker::Send: the size of dyn std::ops::FnOnce() + std::marker::Send cannot be statically determined
   --> C:\Users\vsaas\.cargo\registry\src\github.com-1ecc6299db9ec823\crossbeam-utils-0.8.5\src\thread.rs:449:44`

I thought I saw also some issues with `1.39`, but that was maybe related to an old Cargo.lock file.
Now the lowest working version seems to be `1.36`, I changed it to that.

src/decoder/lossless.rs
```diff
@@ -212,7 +212,7 @@ fn convert_to_u8(frame: &FrameInfo, data: Vec<u16>) -> Vec<u8> {
         data.iter().map(|x| *x as u8).collect()
     } else {
         // we use big endian to conform with PNG
```
fintelia, line 262: nit: please update this comment too

HeroicKatora (approved):
Regarding version changes we'll do a big bump to `1.48` soon. I'm marking this as approved. There may be some other changes in the next few days and depending on order they will create a merge conflict with this one. But don't worry, me or another maintainer should take care of them—just leave the fork open and we should have push access to that specific branch.

PR #196, opened 2021-09-23
Move most uses of the std crate to core and alloc

linkmauve:
`std::io` and `std::error::Error` are the two remaining uses of `std` preventing this crate from building in a `#![no_std]` context.

I’ve tried reimplementing just these parts of `std` in this crate, but it was obviously not a good way forward, and I don’t know what could be a way forward here.  In the meantime, these changes don’t cause any issue, and will be useful for the next person trying to implement `no_std`!

`f32::ceil()` and `f32::fract()` are two other functions we use which are available only in `std`, I’m not too sure why.


HeroicKatora:
@fintelia Your judgment, should we do the same as in the other crates—move to `0.2` and bump to `1.48`? Note that we'll need to do a breaking release in any case after #193 is merged.

HeroicKatora (approved):

PR #204, opened 2021-10-20
Fix bogus check for icc markers

vstroebel:
This don't change anything in the behavior, because 0 is always <256 and the >=256 case is always true in the first condition of the loop below (256 is always larger than what `chunk.num_markers: u8` can represent). 

It would be better to emit a warning in case of invalid ICC profile data but the encoder currently lacks a way to do this...


vstroebel:
> Good find! Do you happen to have a reproduction/regression test that is usable (from a license perspective)?

Because the original code doesn't result in any real bug there isn't any reproducible test possible. But I've added some tests that check certain aspects of chunk handling.

The test images contain only a black pixel with some fake ICC data.

PR #209, opened 2021-12-09
Lossless spectral selection

HeroicKatora:
Fixes an issue where images utilizing some spectral selection patterns would get rejected, due to also requiring being valid for lossless decoding. However, in that case the parameter has a different meaning and must be in a list of enumerated values which must not be enforced for non-lossless cases. This was noticed in the `image` test-suite so the relevant image has been imported.


fintelia (approved):
Looks plausible, though I don't know JPEG well enough to say whether it is correct or not

Issue #202, opened 2021-10-12
Speed is half of normal jpeg decoders (on ARM)


fzyzcjy:
Hi thanks for the lib! However, the speed of this library seems to be half of normal jpeg decoders. When running on ARM, normal decoder uses 600ms to decode a 4000x3000 color JPEG, while this lib uses 1100ms.

veluca93:
@HeroicKatora I have noticed that NEON intrinsics are currently only available on nightly. I can think of 3 options for adding SIMD support on NEON, and I'm not sure which is the best.

- Use an external `asm` file that contains the NEON-specific code. This would work, but it's a huge pain (among other things, one needs to care about calling conventions and the like...)
- Use an external `c` file containing functions implemented with intrinsics. I believe this is entirely superior to the `asm` approach.
- Only support NEON on nightly. By far the easiest option, and saves us the pain of writing the CPU detection code if we want to support arm and not just aarch64...

What are your thoughts?

fintelia:
How likely is it that the NOEN intrinsics will be stabilized in their current form? If it is just a matter of waiting for them to reach stable, then I think that's probably the best option. We'd just leave it behind a feature flag until they were available on stable at our MSRV.

What worries more is the testing/maintenance strategy given that the CI and most developers' machines are currently x86. (Not a deal-breaker mind you, just something to figure out so we don't find ourselves scrambling to track down an ARM machine because we need to test an urgent patch or something).

HeroicKatora:
Regarding NEON, I wonder if it were possible to convince LLVM to auto-vectorize enough and in particular utilizing instructions. We can have specialized methods with explicit `target_feature(enable = "…")` annotations. For example, consider and compare the instruction output here:

```rust
#[target_feature(enable = "avx2")]
pub unsafe fn sum_sse3(vec: &[u32]) -> u32 {
    let mut n = 0u32;
    for &i in vec {
        n = n.wrapping_add(i);
    }
    n
}

pub fn sum_me(vec: &[u32]) -> u32 {
    let mut n = 0u32;
    for &i in vec {
        n = n.wrapping_add(i);
    }
    n
}
```

[playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=ad6589aa8a8c9430ac51eaa9b761c40a)

HeroicKatora:
Ah well, looks like the feature names and annotation for ARM aren't stable either. <https://rustc.godbolt.org/z/bcr59rTaa>
[...]

Issue #79, opened 2017-09-13
Consider accelerated IDCT?


nathanaeljones:
I would imagine that with IDCT accelerated, Jpeg->YCbCr unscaled planar performance should meet/exceed libjpeg-turbo (it hasn't merged these yet - I think these handle multiple blocks). 

https://github.com/huaqiangwang/libjpeg-turbo/blob/avx2-dev/simd/jidctflt-avx2-64.asm
https://github.com/huaqiangwang/libjpeg-turbo/blob/avx2-dev/simd/jfdctflt-avx2-64.asm
https://github.com/huaqiangwang/libjpeg-turbo/blob/avx2-dev/simd/jfdctmflt-avx2-64.asm

kaksmet:
Patches for SIMD acceleration are welcome, I don't have the time to implement it myself.

lilith:
Is this something you would be able to delegate or work on with additional funding?

HeroicKatora:
For the forseeable future the project is mainly limited by dev head count/time. Delegation is not a problem as long as the changes are properly licensed. Porting libjpeg-turbo's SIMD implementation, which seems to be zlib licensed, sounds reasonable. We'd have to figure out how best to test the combination of simd/no-simd and the rest of the code but I don't see any blocking concerns there either.

lilith:
I'm considering using this project in https://github.com/imazen/imageflow as the default jpeg decoder, but I expect I will need to expand things like color profile/exif support and add assembly and C to the build to optimize performance-critical bits. I'd like to make sure these changes align with the project direction before investing. 

Ideally, I'd like to locate someone who can work on this codec full-time for a while and bring it up to parity with libjpeg-turbo (which is not usable in Rust due to setjmp usage).

HeroicKatora:
> Ideally, I'd like to locate someone who can work on this codec full-time for a while and bring it up to parity with libjpeg-turbo (which is not usable in Rust due to setjmp usage).

That would be _awesome_. As for governance, this crate in particular was previously a separate entity. It moved to the `image-rs` organization to maintain and we're open to collaborators and I have no problem with another developer in a lead role per-se.¹

[...]

PR #211, opened 2021-12-16
SSSE3 implementations of 8x8 IDCT and YCbCr conversion

veluca93:
Hi folks!

This PR is more of a proof of concept / demo than an actual PR - I realize there's probably a few things to discuss/agree on/do before merging this in, but I figured a benchmark was the best way to get the discussion started ;). See also #146 for a previous take on this.

This PR contains two SSSE3 implementations of respectively 8x8 IDCT and YCbCr conversion.
Why SSSE3? Two reasons: it is extremely common nowadays (99.25% of x86 PCs according to the Steam Hardware survey), while offering some operations (in particular blends) that provide some significant speedups over SSE3 or SSE2. On the other hand, the next "useful" upgrade in x86 SIMD would be AVX2, which wouldn't help that much (in particular it wouldn't help IDCT) and is not that common (85.98%).

From the benchmark, you can see speedups (with a single thread) of a factor of approximately 2x in the common case of a 444 YCbCr image.

While I didn't (yet!) implement this for ARM NEON, I expect the speedup there to be similar, likely resolving #202. See also #79.

```
decode a 512x512 JPEG   time:   [1.7992 ms 1.8000 ms 1.8009 ms]                                   
                        change: [-49.653% -49.613% -49.574%] (p = 0.00 < 0.05)
                        Performance has improved.

decode a 512x512 progressive JPEG                                                                             
                        time:   [4.5663 ms 4.5848 ms 4.6057 ms]
                        change: [-29.470% -28.997% -28.514%] (p = 0.00 < 0.05)
                        Performance has improved.

decode a 512x512 grayscale JPEG                                                                            
                        time:   [780.95 us 782.16 us 783.31 us]
                        change: [-36.512% -36.366% -36.222%] (p = 0.00 < 0.05)
                        Performance has improved.


decode a 2268x1512 JPEG time:   [22.317 ms 22.339 ms 22.362 ms]                                    
                        change: [-52.237% -52.149% -52.073%] (p = 0.00 < 0.05)
                        Performance has improved.
```

This brings me to the open questions I have related to this PR:
- Unfortunately, different rounding produces slightly different results in the SIMD code, although I am fairly confident this is still within the bounds of conformance as defined in the JPEG standard ("DCTs back to the same coefficients"). I'm not sure if we'd want to make sure that the non-SIMD implementation matches the SIMD one.
- The SIMD implementation doesn't handle all the values that the non-SIMD one does -- in particular it doesn't handle i16::MAX, although to my understanding such values can never appear in a JPEG bitstream.
- As this code uses hand-crafted intrinsics, it makes use of two `unsafe` functions. I'm confident they are safe (they just do arithmetic, swizzle, and loads/store to arrays that are bounds-checked in advance), but it may be a good idea to put the SIMD code in a separate crate, in separate files and/or behind a feature flag.
- Should we also SIMDfy other things? What comes to mind are other color conversions, smaller (4x4 and 2x2) DCTs and chroma upsampling.

I'd be happy to hear your thoughts on this :)


src/idct.rs
```diff
@@ -25,41 +30,393 @@ pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions
...
+        ),
         _ => panic!("Unsupported IDCT scale {}/8", scale),
     }
 }
 
+macro_rules! idct8_ssse3 {
+    ($data:expr) => {
+        let p2 = $data[2];
+        let p3 = $data[6];
+        let p1 = _mm_mulhrs_epi16(_mm_adds_epi16(p2, p3), _mm_set1_epi16(17734)); // 0.5411961
```
HeroicKatora, line 246: Can you outline these constants to some common table/struct and explain them? I'm not sure I understand how they are derived other than the ones related to the common factors. I notice that [jpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/c23672ce52ae53bd846b555439aa0a070b6d2c07/simd/mips/jsimd.c) uses different—related to the standard ones—and fewer constants. Potentially derive the fixpoint representation from the float as a constant, making this easier to check / modify.

HeroicKatora (commented):
Thank you for this PR :tada: 

Regarding folder structures it would make sense to me to have `src/arch/sse3.rs` for those particular algorithms. This has two purposes:
- `#[allow(unsafe)]` could be scoped to _just_ the arch module where it is generally accepted to be required. This would imply having dispatch code in that module as well.
- It groups code by the specialized knowledge required to read it, understand it, review it, change it.

Regarding the code, the implement indeed looks fine on a first impression. I do not consider it to be necessary to aim for 100% reproducability of the non-SIMD code *iff* we make it some flag that can be turned off. It should at least be possible to enforce consistent behavior between systems from the deterministic decoder.

There's also some questions regarding style in the code itself. See below.

src/idct.rs
```diff
@@ -25,41 +30,393 @@ pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions
...
+            coefficients,
+            quantization_table,
+            output_linestride,
+            output,
+        ),
         _ => panic!("Unsupported IDCT scale {}/8", scale),
     }
 }
 
+macro_rules! idct8_ssse3 {
```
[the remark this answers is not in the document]
veluca93: I had tried lambdas and they resulted in pretty horrendous codegen, but it may be the case that separate functions don't have this issue - I'll try.

tests/reftest/mod.rs
```diff
@@ -96,8 +96,7 @@ fn reftest_decoder<T: std::io::Read>(mut decoder: jpeg::Decoder<T>, path: &Path,
         let diff = (a as isize - b as isize).abs();
         max_diff = cmp::max(diff, max_diff);
 
-        // FIXME: Only a diff of 1 should be allowed?
-        if (info.coding_process != jpeg::CodingProcess::Lossless && diff <= 2) || diff == 0  {
+        if (info.coding_process != jpeg::CodingProcess::Lossless && diff <= 3) || diff == 0  {
```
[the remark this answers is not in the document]
veluca93: The SIMD implementation should be pretty much equivalent to the non-SIMD implementation, except for the number of fractional bits used (4 instead of 12, AFAIU) to fit in a 16-bit integer.

How did you produce the "reference" PNGs? Given how under-defined JPEG decoding is, I wouldn't be surprised at all if libjpeg-turbo produces results that are off from the "ideal" values in slightly different ways...

src/idct.rs
veluca93: The general idea here is that x86 has a fast "fixed point multiplication by [-1, 1) constant represented with 15 mantissa bits" operation (FWIW, so does NEON). So the constants here are just the fractional part of the constants in the non-SIMD implementation multiplied by `1<<15` and rounded to the nearest int. Maybe I should put that in a comment ;)

veluca93:
> Thank you for this PR tada
> 
> Regarding folder structures it would make sense to me to have `src/arch/sse3.rs` for those particular algorithms. This has two purposes:
> 
> * `#[allow(unsafe)]` could be scoped to _just_ the arch module where it is generally accepted to be required. This would imply having dispatch code in that module as well.
> * It groups code by the specialized knowledge required to read it, understand it, review it, change it.
> 
> Regarding the code, the implement indeed looks fine on a first impression. I do not consider it to be necessary to aim for 100% reproducability of the non-SIMD code _iff_ we make it some flag that can be turned off. It should at least be possible to enforce consistent behavior between systems from the deterministic decoder.
> 
> There's also some questions regarding style in the code itself. See below.

Replied to some higher-level comments while I deal with the other ones :)

tests/reftest/mod.rs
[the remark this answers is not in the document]
HeroicKatora: Via `convert`, so image magick. Except I'm not 100% sure about all of the setings used in those calls.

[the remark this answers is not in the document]
veluca93: By default imagemagick uses the system libjpeg, usually libjpeg-turbo, which by default uses SIMD - so this likely ended up running the avx2 algorithm of libjpeg-turbo, which is a (different...) 8x8 16-bit IDCT. I'm not actually sure how to convince libjpeg-turbo to use the floating point IDCT...

Either way, IIRC the "official" definition of conformance is "if you DCT the pixels back, before applying any colour transforms, you stay within the quantization buckets of the encoded coefficients", so a bound of 3 here is likely fine :D

src/idct.rs
```diff
@@ -372,70 +744,56 @@ fn stbi_fsh(x: Wrapping<i32>) -> Wrapping<i32> {
...
-        -6, -13, -9, -1, -2, -1, 0, 0,
-        -4, 0, -1, -2, 0, 0, 0, 0,
-        3, 0, 0, 0, 0, 0, 0, 0,
-        -3, -2, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0,
-        0, 0, 0, 0, 0, 0, 0, 0];
+        -14, -39, 58, -2, 3, 3, 0, 1, 11, 27, 4, -3, 3, 0, 1, 0, -6, -13, -9, -1, -2, -1, 0, 0, -4,
+        0, -1, -2, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, -3, -2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    ];
```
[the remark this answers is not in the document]
veluca93: Done.

src/idct.rs
```diff
@@ -25,41 +30,393 @@ pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions
...
+#[target_feature(enable = "ssse3")]
+#[allow(unsafe_code)]
+pub unsafe fn dequantize_and_idct_block_8x8_ssse3(
+    coefficients: &[i16],
+    quantization_table: &[u16; 64],
+    output_linestride: usize,
+    output: &mut [u8],
+) {
+    assert!(coefficients.len() >= 64);
+    assert!(output.len() >= output_linestride * 7 + 8);
```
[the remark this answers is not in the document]
veluca93: Changed to `checked_*` operations and added a comment.

src/idct.rs
```diff
@@ -25,41 +30,393 @@ pub(crate) fn choose_idct_size(full_size: Dimensions, requested_size: Dimensions
...
+
+    let mut data = [_mm_setzero_si128(); 8];
+    for i in 0..8 {
+        data[i] = _mm_slli_epi16(
+            _mm_mullo_epi16(
+                _mm_loadu_si128(coefficients.as_ptr().wrapping_add(i * 8) as *const _),
+                _mm_loadu_si128(quantization_table.as_ptr().wrapping_add(i * 8) as *const _),
+            ),
+            SHIFT,
+        );
```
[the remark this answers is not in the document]
veluca93: I tried iterators and they seem to introduce a 1-3% performance regression, so I think I'll stick with pointer arithmetic if you don't mind :)

src/idct.rs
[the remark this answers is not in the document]
veluca93: Moving to a separate function seems to do the trick.

src/idct.rs
veluca93: Added a comment.

veluca93:
> > Thank you for this PR tada
> > Regarding folder structures it would make sense to me to have `src/arch/sse3.rs` for those particular algorithms. This has two purposes:
> > 
> > * `#[allow(unsafe)]` could be scoped to _just_ the arch module where it is generally accepted to be required. This would imply having dispatch code in that module as well.
> > * It groups code by the specialized knowledge required to read it, understand it, review it, change it.
> > 
> > Regarding the code, the implement indeed looks fine on a first impression. I do not consider it to be necessary to aim for 100% reproducability of the non-SIMD code _iff_ we make it some flag that can be turned off. It should at least be possible to enforce consistent behavior between systems from the deterministic decoder.
> > There's also some questions regarding style in the code itself. See below.
> 
> Replied to some higher-level comments while I deal with the other ones :)

I added a enabled-by-default "simd" feature to allow disabling compilation of the SIMD / unsafe code. PTAL :)

Cargo.toml
```diff
@@ -18,9 +18,21 @@ png = "0.16"
 walkdir = "2.0"
 criterion = "0.3"
 
+[profile.bench]
+debug = true
+
+[profile.release]
+debug = true
+
+
```
fintelia, line 27: This is only for development, right?

veluca93: Ups. Fixed.

src/arch/ssse3.rs
```diff
@@ -0,0 +1,261 @@
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[target_feature(enable = "ssse3")]
```
HeroicKatora, line 7: Read up on this and semantics might be different than what is implied by the usage here. This instructs `rustc` to use ssse3 for _its own code gen_ of this function. This is important for llvm's auto-vectorizer and other intrinsics. However, when you're using intrinsics then you're instructing it to place those exact primitive regardless. Thus I think this may be superfluous. It doesn't hurt anyways.

Cargo.toml
```diff
@@ -22,5 +22,10 @@ criterion = "0.3"
...
 harness = false
 
+[[bench]]
+name = "large_image"
+harness = false
+
 [features]
-default = ["rayon"]
+default = ["rayon", "simd"]
+simd = []
```
HeroicKatora, line 31: I reconsidered the style of the flag here, and maybe it shouldn't be a compile time flag. The arguments for a compile time flag are in general, size of codegen/compile time and compatibility. Neither of those should be an issue. The functions are monomorphized and using the intrinsics is valid for any `x86` platform—it's just that the function must stay unreachable on platforms that do not support _executing_ such intrinsics.

There are downsides to a `feature` flag in particular as well. It is not orthogonal between users. Indeed, if any dependent enables usage of `simd` then _all_ the dependent in the tree with compatible versions will get the feature (and thus run different code). It becomes more complex to include the library because anyone must now make a choice to expose this feature or not. `rayon` in particular had been made a feature to support `wasm` deployment in those versions where multi-threading was strict not possible and any mention of `rayon` lead to compilation failure.

Thus I would suggest making a _runtime_ feature selection and add a 'flag'  to the `Decoder` to toggle it on/off. The default may as well be on due to not having a _strict_ conformance test suite, but I'm open to input on that decision.

src/arch/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+                );
+                return;
+            }
+        }
+    }
+    unreachable!("No arch-specific IDCT available");
+}
+
+/// Returns true if an arch-specific IDCT is avaliable, false otherwise.
+pub fn has_arch_specific_idct() -> bool {
```
HeroicKatora, line 49: In the spirit of 'parse, don't validate', I would return an optional, unsafe function pointer here instead of a bool. This places the detection code much closer to the specific requirements that the unsafe execution requires. It's easier to discern that we return (and execute) an `sss3` function only when `is_x86_feature_detected` is true.

HeroicKatora (commented):
I like the new structure better. Thanks for integrating all of the feedback, this is quite a new kind of addition to any of the libraries (`png` mostly relies on auto-vectorization, if you're looking for a task afterwards :smirk:) and thus I'd like to find a solution that we can use as a template for similar additions otherwhere.

veluca93:
> I like the new structure better. Thanks for integrating all of the feedback, this is quite a new kind of addition to any of the libraries (`png` mostly relies on auto-vectorization, if you're looking for a task afterwards smirk) and thus I'd like to find a solution that we can use as a template for similar additions otherwhere.

FWIW I doubt SIMD in `png` decoders can help much :D (perhaps in the encoder...)

src/arch/ssse3.rs
veluca93: So I don't *really* know what is going on here, but disabling target_feature seems to **significantly** slow things down, so I'd rather leave it here :D

Cargo.toml
veluca93: So the main benefit I can see of a feature flag is: "if you disable the simd feature, there is no unsafe code at all" (maybe that can even be made explicit with a `#[forbid]`? Not 100% sure yet of what you can do for this)

As far as I'm aware, no other image format library (at least the ones I'm familiar with - libjxl, libjpeg-turbo and libpng up to a point) has a runtime flag to decide between SIMD or not, so there's that too...

src/arch/mod.rs
veluca93: What would this do when there's no SIMD option? I guess return a None?

This could work, but I'm not sure if the extra indirection might cause some speed problems. Probably not, but I'll have to test.

src/arch/ssse3.rs
HeroicKatora: Interesting. I'd really like to find out why that is but indeed, then let's leave it here, and that a task for myself :joy:

Cargo.toml
HeroicKatora: If that is the intent then it should be `#![cfg_attr(not(feature = "simd"), forbid(unsafe_code)]` (quite the mouthful).

Regarding the flag, I wouldn't call it `SIMD` but rather `platform_compatible`. Maybe even `strict`, referencing the attribute on glsl code. That is the intent is not the slow things down or speed them up but rather to make the result independent of platform—i.e. the decoding should be a pure function.

If we do _not_ have such a mode then this can effectively not be used in browsers, or it would allow to fingerprint the user's processor architecture quite clearly by decoding certain images to canvas.

src/arch/mod.rs
HeroicKatora: Returning a function point (`fn(…) -> …`)–or it should be more aptly named function reference as it must be non-null—is not double indirection. That's the exact type of the symbol of a function.

HeroicKatora:
> FWIW I doubt SIMD in png decoders can help much :D (perhaps in the encoder...)

There's a whole adler32 stage that's going to be simd-ified soon for some speed gains. Plus it seems that the main zlib (through `miniz_oxide`) is not simdified despite `zlib-ng` demonstrating how platform-specific code might show speed gains over the similarly written `zlib`. And then the actual color stage involve some line-by-line filtering, permutation of bytes, etc. There's always code that can be sped up by careful SIMD. The big question is whether LLVM has already done so implicitly but then it's most likely a 'generic' variant based on avx or w/e is in the default set of extensions of modern x86-64 ABI.

src/arch/mod.rs
veluca93: Done, no slowdown I could measure :)

Cargo.toml
veluca93: I'm not sure browsers care that much about that - I believe you can at least get the GPU *model* with JS. You can probably also do this in a couple of other ways, i.e. wasm-SIMD exists.

Anyway, renamed it to `platform_independent` (and flipped the logic). PTAL.

veluca93:
> > FWIW I doubt SIMD in png decoders can help much :D (perhaps in the encoder...)
> 
> There's a whole adler32 stage that's going to be simd-ified soon for some speed gains. Plus it seems that the main zlib (through `miniz_oxide`) is not simdified despite `zlib-ng` demonstrating how platform-specific code might show speed gains over the similarly written `zlib`. And then the actual color stage involve some line-by-line filtering, permutation of bytes, etc. There's always code that can be sped up by careful SIMD. The big question is whether LLVM has already done so implicitly but then it's most likely a 'generic' variant based on avx or w/e is in the default set of extensions of modern x86-64 ABI.

In my experience, most of the gains from manual SIMDfication would be in encoding for lz77, not in decoding - decoding is simple enough that compilers do a reasonable job about it. Although I might be wrong ;) (also probably adler32)

As for other things in `png`: line-by-line filtering in the decoder would be rather tricky to SIMDfy (except perhaps across the channels, but then you don't gain that much - IIRC libpng does that). Permutation of bytes would benefit though.

veluca93:
@HeroicKatora: Any clue why the CI would fail on beta and on beta only?

fintelia:
Not sure why that's happening, but I tried rerunning the CI and the same test failed. It is probably worth trying to replicate locally

veluca93:
> Not sure why that's happening, but I tried rerunning the CI and the same test failed. It is probably worth trying to replicate locally



> Not sure why that's happening, but I tried rerunning the CI and the same test failed. It is probably worth trying to replicate locally

I can reproduce locally both on beta and nightly, and with and without rayon -- albeit on a different file...

This is looking like it will be a fun investigation :)

Cargo.toml
HeroicKatora: Seems fine. In any case, at least Tor cares because when I try to get any `WEBGL_debug_renderer_info` then the result is null—thus no exact information on any hardware is available. I would really like if more browsers cared but alas that's the world when Browser vendors themselves _want_ to fingerprint their users.

veluca93:
> > Not sure why that's happening, but I tried rerunning the CI and the same test failed. It is probably worth trying to replicate locally
> 
> I can reproduce locally both on beta and nightly, and with and without rayon -- albeit on a different file...
> 
> This is looking like it will be a fun investigation :)

I was right - somehow, between stable and beta some intrinsics managed to produce slightly different effects, so I modified the code not to use those intrinsics :)

HeroicKatora:
> I was right - somehow, between stable and beta some intrinsics managed to produce slightly different effects

Have you checked [rust-lang/rust](https://github.com/rust-lang/rust) for issues? This sounds like a serious codegen bug if the effect of (unsafe) instructions is incorrect.

veluca93:
> > I was right - somehow, between stable and beta some intrinsics managed to produce slightly different effects
> 
> Have you checked [rust-lang/rust](https://github.com/rust-lang/rust) for issues? This sounds like a serious codegen bug if the effect of (unsafe) instructions is incorrect.

Yup, there is https://github.com/rust-lang/rust/issues/84042 :)

src/arch/ssse3.rs
```diff
@@ -0,0 +1,262 @@
...
+pub unsafe fn color_convert_line_ycbcr(y: &[u8], cb: &[u8], cr: &[u8], output: &mut [u8]) -> usize {
+    assert!(output.len() % 3 == 0);
+    let num = output.len() / 3;
+    assert!(num <= y.len());
+    assert!(num <= cb.len());
+    assert!(num <= cr.len());
+    // _mm_loadu_si64 seems to be confused between stable and beta on whether it loads on the low
+    // or the upper half of the vector. To circumvent the issue, we use _mm_loadu_si128, and skip
+    // one vector to avoid loads from outside accessible memory.
+    let num_vecs = (num / 8).saturating_sub(1);
```
HeroicKatora, line 187: Sounds nice in theory but what happens to the last vector now? There doesn't appear to be a fixup at the end. Also, the path to properly fixing this issue is unclear to me. As far as I understand the correct instruction _could_ be used starting with Rust `1.58`. However, doing so would clearly lead to incorrect code in Rust `<1.57`. Clearly, this version number should be added in the comment here somewhere so that will be picked up for the msrv description if we decide to go back to the 'proper' intrinsic.

veluca93: That's not an issue - leftover pixels are handled by the non-SIMD code (and they'd need to be either way for images that are not multiples of 8/16 pixels). So this change only increases that number a little bit. I'll update the comment to explicitly mention the Rust version.

HeroicKatora: Ah I see, the change has also implicitly influence the number of handled pixels that's being returned. A smarter implementation than my quick review of the PR :sweat_smile: Very good.

HeroicKatora (commented):
LGTM. Can you strew in just a few more comments on the _approach_ and constants, rather than the specific implementation?

src/arch/ssse3.rs
```diff
@@ -0,0 +1,263 @@
...
+    idct8(&mut data);
+    transpose8(&mut data);
+
+    for i in 0..8 {
+        let mut buf = [0u8; 16];
+        _mm_storeu_si128(
+            buf.as_mut_ptr() as *mut _,
+            _mm_packus_epi16(
+                _mm_srai_epi16(
+                    _mm_adds_epi16(data[i], _mm_set1_epi16(257 << (SHIFT + 2))),
```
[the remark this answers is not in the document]
veluca93: I added some comments to this part, and in general to the whole file. PTAL.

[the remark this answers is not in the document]
HeroicKatora: Perfect, thank you.

HeroicKatora (approved):
LGTM.

veluca93:
Hi!
I was thinking of doing the aarch64/arm version, but would it be possible to merge this PR first?

src/arch/ssse3.rs
```diff
@@ -0,0 +1,289 @@
...
+    data[4] = _mm_unpacklo_epi64(d0123hl, d4567hl);
+    data[5] = _mm_unpackhi_epi64(d0123hl, d4567hl);
+    data[6] = _mm_unpacklo_epi64(d0123hh, d4567hh);
+    data[7] = _mm_unpackhi_epi64(d0123hh, d4567hh);
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[target_feature(enable = "ssse3")]
+pub unsafe fn dequantize_and_idct_block_8x8(
+    coefficients: &[i16],
```
paolobarbolini, line 125: Drive-by review: just like #215 replaced some slices with arrays it would make sense to do the same here

HeroicKatora:
Some formatting changes in another branch had caused a conflict, I fixed that formatting in a merge.

PR #213, opened 2022-01-07
Remove feature gate on the std feature

paolobarbolini:
This improves the decode performance to up to 40%. Am I missing something? I wouldn't expect something this simple to add such a big change :sweat_smile:.
The feature gate was added by #196, but no `std` feature is present in the Cargo.toml, so the multithreaded worker would never get used.

### DIFF
--- a/src/worker/mod.rs
+++ b/src/worker/mod.rs
@@ -1,10 +1,9 @@
 mod immediate;
-#[cfg(feature = "std")]
 mod multithreaded;
 
-#[cfg(all(feature = "std", not(any(target_arch = "asmjs", target_arch = "wasm32"))))]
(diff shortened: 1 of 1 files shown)


paolobarbolini:
Force pushed to change email

fintelia:
The alternative would be to add the `std` feature to Cargo.toml, but I think your way is a better option seeing as #196 points out that the crate cannot actually work without the standard library.

paolobarbolini:
> The alternative would be to add the `std` feature to Cargo.toml, but I think your way is a better option seeing as #196 points out that the crate cannot actually work without the standard library.

It could be added in 0.3.0

PR #220, opened 2022-02-12
Release notes and version bump to 0.2.2

HeroicKatora:


HeroicKatora:
Thanks for all the change note and publish reviews, by the way. This may be the first one where I didn't typo anything in the first try, I'm glad you caught all the other ones.

PR #221, opened 2022-02-13
Use NEON instructions on aarch64.

### LINKED ISSUE #202: Speed is half of normal jpeg decoders (on ARM)

veluca93:
The code is a direct translation of the sse3 code for x86, and provides
approximately a 0.67x decrease of image decoding time on a Samsung A51.

Fixes #202, or at least improves the situation.

```
decode a 2268x1512 JPEG time:   [83.619 ms 85.692 ms 87.848 ms]

decode a 2268x1512 JPEG time:   [56.209 ms 57.019 ms 57.876 ms]
                        change: [-35.318% -33.460% -31.535%] (p = 0.00 < 0.05)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,3 +29,5 @@ harness = false
 [features]
 default = ["rayon"]
 platform_independent = []
+nightly_aarch64_neon = []
+
--- a/src/arch/mod.rs
+++ b/src/arch/mod.rs
@@ -1,5 +1,6 @@
 #![allow(unsafe_code)]
 
+mod neon;
 mod ssse3;
 
 #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -29,3 +29,5 @@ harness = false
 [features]
 default = ["rayon"]
 platform_independent = []
+aarch64_neon = []
```
HeroicKatora, line 32: For clarity, this is only a feature because it will depend on a nightly compiler. There's no reason to treat it differently, otherwise. Correct? In this case I would rename the feature to `nightly_aarch64_neon`.

src/arch/mod.rs
```diff
@@ -16,18 +17,30 @@ pub fn get_color_convert_line_ycbcr() -> Option<unsafe fn(&[u8], &[u8], &[u8], &
             return Some(ssse3::color_convert_line_ycbcr);
         }
     }
+    // Runtime detection is not needed on aarch64.
+    #[cfg(all(feature = "aarch64_neon", target_arch = "aarch64"))]
+    {
+        return Some(neon::color_convert_line_ycbcr);
```
HeroicKatora, line 23: Is there no  [`is_aarch64_feature_detected`](https://doc.rust-lang.org/stable/std/macro.is_aarch64_feature_detected.html) code for this?

HeroicKatora (commented):
Overall clean, very similar to the x86 code iirc (if i read correctly). Is there any reason why this deviates from the architecture of x86-related arch module—dynamic feature detection and no dedicated compile time feature, e.g. different stabilization plans?

Cargo.toml
veluca93: Done.

src/arch/mod.rs
veluca93: aarch64 makes neon mandatory, so such a function would always return true for NEON :)

HeroicKatora: I see :joy:

veluca93:
Should we add a CI run on a emulated aarch64? If so, in this PR or in another?

veluca93:
> If we can, that'd be perfect. It can be another PR though I'd prefer adding it before this one then.

Done and new feature added to the CI.

HeroicKatora (approved):
LGTM then.

PR #226, opened 2022-02-18
Add no_std support

### LINKED ISSUE #199: Make this crate usable in a no_std environment
#196 removed most uses of `std`, but as said there `std::io` and `std::error::Error` are two remaining uses of `std` with no `no_std` replacement.  Here is the tracking issue: https://github.com/rust-lang/rust/issues/48331

This issue is about adding an alternative API which wouldn’t use the `io::Read` trait for reading JPEG data, so that it could be used in embedded environments.

vstroebel:
This is my attempt to make this crate usable on no_std.

I've splitted this changes into several commits to make it easier to review:

1. Introduce trait `JpegRead` as an replacement for std::io::Read. On `std` `JpegRead` is implemented for every `std::io:::Read` which should allow users to use every type implementing `Read`, too.
2. Some `use` relate changes that #196 didn't cover
3. Integer only replacement for f32::ceil based code that isn't available in ´core´
4. An ugly replacement for f32::fract. Maybe someone has a better idea to implement this for no_std.
5. An integration test that works on no_std. The integration test suite heavily depends on file system ops and the png crate. Having at least one test that works on no_std is better than nothing....
6. Add a `std` feature and use it as a marker for std/no_std. 
7. Change github action to test std and no_std code

This should fix #199


src/reader.rs
```diff
@@ -0,0 +1,83 @@
...
+#[cfg(not(feature = "std"))]
+impl JpegRead for &[u8] {
+    fn read_exact(&mut self, buf: &mut [u8]) -> Result<(), Error> {
+        if buf.len() <= self.len() {
+            let (data, remaining) = self.split_at(buf.len());
+            buf.copy_from_slice(data);
+            *self = remaining;
+            Ok(())
+        } else {
+            panic!();
```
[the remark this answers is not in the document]
HeroicKatora: And now that it existed once, my desire for much more extensive test coverage of all the impls has grown quite a lot.

src/upsampler.rs
```diff
@@ -176,7 +183,7 @@ impl Upsample for UpsamplerH1V2 {
         let row_near = row as f32 / 2.0;
         // If row_near's fractional is 0.0 we want row_far to be the previous row and if it's 0.5 we
         // want it to be the next row.
-        let row_far = (row_near + row_near.fract() * 3.0 - 0.25).min((input_height - 1) as f32);
+        let row_far = (row_near + fract(row_near) * 3.0 - 0.25).min((input_height - 1) as f32);
```
HeroicKatora, line 186: Can we not just use integer fixed-point math here?

tests/lib.rs
```diff
@@ -1,134 +1,145 @@
...
+#[cfg(feature = "std")]
 mod reftest;
 
-#[test]
-fn read_info() {
-    let path = Path::new("tests").join("reftest").join("images").join("mozilla").join("jpg-progressive.jpg");
+mod no_std;
+
+#[cfg(feature = "std")]
+mod std_tests {
```
HeroicKatora, line 15: I wouldn't mind if the whole module were moved to a separate file here.

HeroicKatora (commented):
Opinion: This affects too much code, too visibly. Infecting and changing all _public_ bounds is just one uncomfortable change that _most_ users won't be positively affected by. The structure of this implementation forces us to do all-or-nothing, however I don't think this ready for merging. I'd like at least a significantly better test coverage of individual (and somewhat awkward) floating point workarounds, and that they not be present in the usual configuratio. Some more detailed concers with specific implementation choices are listed below.

Overall, I would much rather we have crate that addresses those concerns separately. One that can be shared across multiple of `image-rs` crates for no_std compat (`png`, `tiff` would be good starts) and provides `Seek` maybe as well. With focussed test coverage, orthogonality of design. There are plenty but none I'm aware of have been too convincing regarding the SemVer / compatibility problems briefly mentioned below.

Issue #227, opened 2022-02-21
rayon feature causes deadlocks when used inside of an existing rayon threadpool


awused:
When loading a jpeg with the rayon feature enabled the calling thread does no work of its own and only blocks until the work is completed on other threads, waiting for mpsc messages to come through. If the task is already running from inside the context of a threadpool this will block one of those threads until the other threads can process the actions. If an application is using a rayon threadpool - even just the global default rayon pool - to parallelize the loading of images it's possible for enough jpegs to start loading at once to exhaust the pool and deadlock themselves.

I believe refactoring it to use rayon::in_place_scope would work, but a more expedient option is probably to just have a dedicated rayon threadpool for decoding. If nothing else this should be documented and maybe made non-default.

This might apply to other image decoders if any use rayon, I've not inspected their code, but jpegs are especially common.

HeroicKatora:
If I understand, then the situation is as follows:

```
rayon-Pool
|-----|
|  …  |
| Jpg | --> { decode_internal } ----------\
|  …  |                   blocked on multithreaded.rs#L54
|  …  |                                    v
|-----|<--blocked-on free worker--| Component0 | Component1 | Component2 |

```

Is it possible to write a reliable test/reproduction for this? The best fix for this, that I cant think of, would be to allow the current thread to complete each work item, alternatively to of the block on `recv`. This would allow rayon to silently degrade to single-threaded use by having the Jpg-Worker complete the components itself. However, I'm not sure how to achieve this with the API. The current parallelization strategy is also somewhat crude in the first place.

HeroicKatora:
Minimized reproduction:

```rust
    let pool = rayon::ThreadPoolBuilder::new()
        .num_threads(1)
        .build()
        .unwrap();
    pool.install(|| {
        let mut decoder = Decoder::new(File::open(&path).unwrap());
        let _ = decoder.decode().unwrap();
        let _ = decoder.info().unwrap();
    });
```

HeroicKatora:
[...]

PR #230, opened 2022-02-24
Restructure multithreading

HeroicKatora:
Redefine the Worker trait, which allows the chosen worker to create a
new scope. This is relevant for a newly created (and installed) rayon
worker.

The rayon thread pool is now local to the decoding step. This fixes an
issue where improper task scheduling would deadlock decoding. It's not
clear how the intended task scheduling can be reliably achieved without
the guarantee of having at least a second, free, worker thread.

Closes: #227

### DIFF
--- a/src/worker/mod.rs
+++ b/src/worker/mod.rs
@@ -1,11 +1,6 @@
 mod immediate;
 mod multithreaded;
 
-#[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
-pub use self::multithreaded::MultiThreadedWorker as PlatformWorker;
-#[cfg(any(target_arch = "asmjs", target_arch = "wasm32"))]
-pub use self::immediate::ImmediateWorker as PlatformWorker;
-
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 use crate::error::Result;
@@ -17,9 +12,22 @@ pub struct RowData {
     pub quantization_table: Arc<[u16; 64]>,
 }
 
-pub trait Worker: Sized {
-    fn new() -> Result<Self>;
+pub trait Worker {
     fn start(&mut self, row_data: RowData) -> Result<()>;
     fn append_row(&mut self, row: (usize, Vec<i16>)) -> Result<()>;
     fn get_result(&mut self, index: usize) -> Result<Vec<u8>>;
+}
+
+pub enum PreferWorkerKind {
+    Immediate,
+    Multithreaded,
+}
+
+/// Execute something with a worker system.
+pub fn with_worker<T>(prefer: PreferWorkerKind, f: impl FnOnce(&mut dyn Worker) -> T) -> T {
+    match prefer {
+        #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
+        PreferWorkerKind::Multithreaded => self::multithreaded::with_multithreading(f),
+        _ => self::immediate::with_immediate(f),
+    }
 }
--- a/src/worker/mod.rs
+++ b/src/worker/mod.rs
@@ -1,5 +1,7 @@
 mod immediate;
 mod multithreaded;
+#[cfg(feature = "rayon")]
+mod rayon;
 
 use alloc::sync::Arc;
 use alloc::vec::Vec;
@@ -16,6 +18,15 @@ pub trait Worker {
     fn start(&mut self, row_data: RowData) -> Result<()>;
     fn append_row(&mut self, row: (usize, Vec<i16>)) -> Result<()>;
     fn get_result(&mut self, index: usize) -> Result<Vec<u8>>;
+    /// Default implementation for spawning multiple tasks.
+    fn append_rows(&mut self, row: &mut dyn Iterator<Item=(usize, Vec<i16>)>)
+        -> Result<()>
+    {
+        for item in row {
+            self.append_row(item)?;
+        }
+        Ok(())
+    }
 }
 
 pub enum PreferWorkerKind {
@@ -26,6 +37,9 @@ pub enum PreferWorkerKind {
 /// Execute something with a worker system.
 pub fn with_worker<T>(prefer: PreferWorkerKind, f: impl FnOnce(&mut dyn Worker) -> T) -> T {
     match prefer {
+        #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
+        #[cfg(feature = "rayon")]
+        PreferWorkerKind::Multithreaded => self::rayon::with_rayon(f),
         #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
         PreferWorkerKind::Multithreaded => self::multithreaded::with_multithreading(f),
         _ => self::immediate::with_immediate(f),
--- a/src/worker/multithreaded.rs
+++ b/src/worker/multithreaded.rs
@@ -11,13 +11,7 @@ use super::{RowData, Worker};
 use super::immediate::ImmediateWorker;
 
 pub fn with_multithreading<T>(f: impl FnOnce(&mut dyn Worker) -> T) -> T {
-    #[cfg(not(feature = "rayon"))]
-    return self::enter_threads(f);
-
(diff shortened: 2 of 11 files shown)


src/worker/multithreaded.rs
```diff
@@ -4,34 +4,48 @@
...
-    }
-    fn start(&mut self, row_data: RowData) -> Result<()> {
+pub struct StdThreadWorker(MpscWorker);
+
+impl MpscWorker {
+    fn start_with(
+        &mut self,
+        row_data: RowData,
+        spawn_worker: impl FnOnce(usize) -> Result<Sender<WorkerMsg>>,
+    ) -> Result<()> {
```
fintelia, line 41: This whole function could just be:
```rust
// if there is no worker thread for this component yet, start one
if self.senders[row_data.index].is_none() {
    self.senders[row_data.index] = Some(spawn_worker_thread(row_data.index)?);
}
self.senders[row_data.index]
    .as_mut()
    .unwrap()
    .send(WorkerMsg::Start(row_data))
    .expect("jpeg-decoder worker thread error");
Ok(())
```

HeroicKatora: It may well be possible to utilize `Option::get_or_insert_with` as well, to remove that explicit `unwrap`. I just didn't want to mix refactor and function change, but I'll take a look.

fintelia: Understood. I was mostly pointing this out because I was confused why the code was so complicated when I read it the first time.

HeroicKatora:
Added _more_ tests, no based off the global pool, to demonstrate/test that this does not make a quantifiable difference. As suggested to add them [here](https://github.com/image-rs/jpeg-decoder/issues/227#issuecomment-1051128315). @awused Did I miss any, do you have more suggestions? You can easily verify that _all_ of the tests hang with the previous implementation.

src/worker/multithreaded.rs
```diff
@@ -40,25 +54,42 @@ impl Worker for MultiThreadedWorker {
...
+        self.0.append_row(row)
+    }
+    fn get_result(&mut self, index: usize) -> Result<Vec<u8>> {
+        self.0.get_result_with(index, collect_worker_thread)
+    }
+}
 
-    spawn(component, move || {
+fn create_worker() -> (Sender<WorkerMsg>, impl FnOnce() + 'static) {
+    let (tx, rx) = mpsc::channel();
```
awused, line 91: The fact that this is being done at all means rayon is being used as a pool of generic spawned threads. I can't see any advantages to using rayon over any other simple threadpool implementation here, which could do it with less overhead.

HeroicKatora: We're not recreating threads, that's still a small advantage. But indeed, you're welcome to create a PR that implements the result passing in a proper manner otherwise. I, franlky, don't see how except by restructuring due to the absence of `JoinHandle` or other futures for tasks?

awused: Any generic threadpool implementation can avoid recreating threads the same way, rayon is a large dependency to pull in when it's only being used to spawn threads and join them at the end. Your existing std multi-threading code is probably marginally more efficient than this code.

HeroicKatora:
Okay, scratch that. This is why it was a draft. How about this: In this `rayon` case, we _always_ `in_place_scope` the main decoder loop. (The architecture should already be there after the rework). Then, instead of creating workers, in `append_row` we run the **Immediate** worker as a new scoped task—this is still single threaded, non-blocking, but I'll get to it. Both `start` and `get_result` will run directly on the immediate worker without any of dubious communication structure of the multithreaded module.

Now the critical part: We modify `Worker::append_row` such that multiple rows _can_ be spawned at the same time. The rayon implementation can then create multiple tasks in its scope, which lets other threads worksteal from us. Not sure about the runtime implications of this scheme. However, this way making the rayon worker  entirely non-blocking might be easier than imagined.

HeroicKatora:
I've given `rayon` usage a complete overhaul. No more blocking at all but `Mutex` to modify the shared result sets.

src/worker/mod.rs
```diff
@@ -17,9 +14,34 @@ pub struct RowData {
...
+/// Execute something with a worker system.
+pub fn with_worker<T>(prefer: PreferWorkerKind, f: impl FnOnce(&mut dyn Worker) -> T) -> T {
+    match prefer {
+        #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
+        #[cfg(feature = "rayon")]
+        PreferWorkerKind::Multithreaded => self::rayon::with_rayon(f),
+        #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
+        PreferWorkerKind::Multithreaded => self::multithreaded::with_multithreading(f),
+        _ => self::immediate::with_immediate(f),
+    }
```
HeroicKatora, line 46: Note: this is not final, but setup to always choose rayon over multithreading over immediate. Note that we may choose _any_ available one at runtime with this style.

src/worker/rayon.rs
```diff
@@ -0,0 +1,117 @@
...
+pub struct Scoped {
+    inner: Mutex<ImmediateWorker>,
+}
+
+pub fn with_rayon<T>(f: impl FnOnce(&mut dyn Worker) -> T) -> T {
+    rayon::in_place_scope(|_| {
+        let inner = ImmediateWorker::new_immediate();
+        f(&mut Scoped { inner: Mutex::new(inner) })
+    })
+}
```
fintelia, line 19: If I'm reading the rayon documentation correctly, this function is equivalent to:
```rust
pub fn with_rayon<T>(f: impl FnOnce(&mut dyn Worker) -> T) -> T {
    let inner = ImmediateWorker::new_immediate();
    f(&mut Scoped { inner: Mutex::new(inner) })
}
```
...which doesn't actually use rayon at all.

fintelia: Specifically this line:
> ...the closure runs on the same thread that calls in_place_scope(). Only work that it spawns runs in the thread pool.

HeroicKatora: Keep in mind this includes work done inside closure `f` (which will be calling `Decoder::decode` etc.). The recursively includes calls to spawn tasks into the thread pool in the `append_row{,s}` impls. The reasoning behind choosing `in_place_scope` specifically:
- We want a scope to guarantee that no more task are running after returning, i.e. cleanup.
- We want `in_place`so as to not require the parameter `f` to have `Sync + 'static` bounds.

fintelia: The problem is that the code ignores the closure argument (a `&rayon::Scope`), which is required to spawn tasks within the scope. See for instance this [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3fe1e5e1fda9660418ed1cb362126f24) which prints the lines out of order

src/worker/rayon.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn get_result(&mut self, index: usize) -> Result<Vec<u8>> {
+        let result = self.inner.get_mut().unwrap().get_result_immediate(index);
+        Ok(result)
+    }
+
+    // Magic sauce, these _may_ run in parallel.
+    fn append_rows(&mut self, iter: &mut dyn Iterator<Item = (usize, Vec<i16>)>) -> Result<()> {
+        rayon::in_place_scope(|scope| {
```
awused, line 156: I think you can drop the mutex everywhere except for the final `write_back` above if you iterate over the set of tasks twice. The only writes in `append_row_locked` are to the `results` array so the locking here and for computing the DCT parameters can be skipped.

Removing the mutex entirely without dropping to unsafe seems possible since the writes don't overlap, but possibly trickier than it's worth. You'd have to benchmark whether keeping track of everything and using slice::split_at_mut would actually be faster than just keeping the lock. If the tasks are sorted, or can be sorted, it'd probably make it easier.

HeroicKatora: Rust lifetimes and borrow checker are amazing when you _really_ use them. We can get rid of mutex entirely, with no unsafe as you said. Tasks are sorted within each component.

PR #236, opened 2022-04-30
Fixing issue with worker on wasm targets

gents83:
With previous code on wasm target it was executing anyway multithreading request

image

### DIFF
--- a/src/worker/mod.rs
+++ b/src/worker/mod.rs
@@ -34,14 +34,18 @@ pub enum PreferWorkerKind {
     Multithreaded,
 }
 
+
 /// Execute something with a worker system.
 pub fn with_worker<T>(prefer: PreferWorkerKind, f: impl FnOnce(&mut dyn Worker) -> T) -> T {
+    #[cfg(any(target_arch = "asmjs", target_arch = "wasm32"))]
+    {
+        return self::immediate::with_immediate(f);
+    }
     match prefer {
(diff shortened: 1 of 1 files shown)


src/worker/mod.rs
```diff
@@ -34,14 +34,18 @@ pub enum PreferWorkerKind {
...
+
 /// Execute something with a worker system.
 pub fn with_worker<T>(prefer: PreferWorkerKind, f: impl FnOnce(&mut dyn Worker) -> T) -> T {
+    #[cfg(any(target_arch = "asmjs", target_arch = "wasm32"))]
+    {
+        return self::immediate::with_immediate(f);
+    }
     match prefer {
-        #[cfg(not(any(target_arch = "asmjs", target_arch = "wasm32")))]
         #[cfg(feature = "rayon")]
```
fintelia, line 45: Would replacing this with `#[cfg(all(not(any(target_arch = "asmjs", target_arch = "wasm32")), feature = "rayon"))]` work instead?

gents83: Yep, it's the same 😊
It is the "double line" cfg the real issue

fintelia: Could we do that instead then? Seems nicer than duplicating the self::immediate::with_immediate() call.

fintelia (approved):

PR #237, opened 2022-05-02
Updated version

gents83:
Update version in order to be updated and used then into image-rs


HeroicKatora:
Oops.. That wasn't intended to happen

PR #239, opened 2022-05-02
Release 0.2.5

HeroicKatora:
Replaces #237 

Since we do releases based on the feature branch itself (not the automatic merge, as it's unnnamed and potentially different) rebased this on master and pushed. Originally, I wanted to push this to the fork but accidentally pushed `master`, which auto-closed the through github. And without an active PR I'm no longer allowed to update the branch / undo that mistake :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## v0.2.5 (2022-05-02)
+
+- Fix single threaded usage in WASM target.
+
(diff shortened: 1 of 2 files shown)


HeroicKatora:
@gents83 sorry for the confusion.

HeroicKatora:
And thank you for the release-notes and metadata bump :)

gents83:
@HeroicKatora may I ask you to integrate it also in image pls?

HeroicKatora:
There's no need to do it explicitly. It's compatible with the current spec (`^0.2.1`) and all it should take is a `cargo update` to renew the dependency. Let me know if there's an issue keeping this from working.

PR #240, opened 2022-05-03
Formatting and isolating rayon feature completely

gents83:
@HeroicKatora this is actually fully separating rayon to make it work on wasm target.
I had other modification in local I didn't integrate in my branch before merging - my bad :(
This will also require to update current release version probably

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -1,3 +1,14 @@
+use crate::error::{Error, Result, UnsupportedFeature};
+use crate::huffman::{fill_default_mjpeg_tables, HuffmanDecoder, HuffmanTable};
+use crate::marker::Marker;
+use crate::parser::{
+    parse_app, parse_com, parse_dht, parse_dqt, parse_dri, parse_sof, parse_sos,
+    AdobeColorTransform, AppData, CodingProcess, Component, Dimensions, EntropyCoding, FrameInfo,
+    IccChunk, ScanInfo,
+};
+use crate::read_u8;
+use crate::upsampler::Upsampler;
+use crate::worker::{with_worker, PreferWorkerKind, RowData, Worker};
 use alloc::borrow::ToOwned;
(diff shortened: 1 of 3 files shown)


src/decoder.rs
```diff
@@ -1174,7 +1180,10 @@ fn compute_image(
     }
 }
 
-#[cfg(feature = "rayon")]
+#[cfg(all(
+    not(any(target_arch = "asmjs", target_arch = "wasm32")),
+    feature = "rayon"
+))]
```
HeroicKatora, line 1186: Could you move this implementation into `worker/rayon.rs`? That would isolate the feature requirement by source location. Also allows fixing the glob import `rayon::prelude::*` and turn it into a proper one.

HeroicKatora (commented):
Is there any way we can test this is CI? My confidence in future changes preserving the wasm-functionality is a bit low. Apart from the nit, looks good.

gents83:
> Is there any way we can test this is CI? My confidence in future changes preserving the wasm-functionality is a bit low. Apart from the nit, looks good.

Mmmm... it'll really depends in what will be added in rayon module little by little, no?
Let's say that the best is to do what you're suggesting - to put all the rayon-related stuff inside rayon.rs - no?

src/decoder.rs
gents83: Is it ok what I did in https://github.com/image-rs/jpeg-decoder/pull/240/commits/7d843f15a19d3354cdf230243b3051dec785b100 ?

HeroicKatora: Yes :+1:

HeroicKatora (approved):
Found a way to test this in CI. So I can confidently say, this solution works and is complete :+1:

HeroicKatora:
Release when the testing as detailed in #241 lands on CI, eta tomorrow.

PR #247, opened 2022-05-14
Sketch: choose worker late

HeroicKatora:
cc: @Shnatsel if you want to try parameter tweaking

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -8,7 +8,7 @@ use crate::parser::{
 };
 use crate::read_u8;
 use crate::upsampler::Upsampler;
-use crate::worker::{compute_image_parallel, with_worker, PreferWorkerKind, RowData, Worker};
+use crate::worker::{compute_image_parallel, PreferWorkerKind, RowData, Worker, WorkerScope};
 use alloc::borrow::ToOwned;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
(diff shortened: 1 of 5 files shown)


src/decoder.rs
```diff
@@ -196,11 +196,31 @@ impl<R: Read> Decoder<R> {
         Some(data)
     }
 
+    /// Heuristic to avoid starting thread, synchronization if we expect a small amount of
+    /// parallelism to be utilized.
+    fn select_worker(frame: &FrameInfo, worker_preference: PreferWorkerKind) -> PreferWorkerKind {
+        // TODO: figure out the exact value.
+        const PARALLELISM_THRESHOLD: u64 = 1024 * 1024;
```
Shnatsel, line 203:
```suggestion
        const PARALLELISM_THRESHOLD: u64 = 128 * 128;
```

Extremely tall images with this area benefit from skipping multithreading (10% speedup) but extremely wide images with this area see no difference.

src/decoder.rs
```diff
@@ -196,11 +196,31 @@ impl<R: Read> Decoder<R> {
...
+
+        match worker_preference {
+            PreferWorkerKind::Immediate => PreferWorkerKind::Immediate,
+            PreferWorkerKind::Multithreaded => {
+                let width: u64 = frame.output_size.width.into();
+                let height: u64 = frame.output_size.width.into();
+                if width * height * frame.components.len() as u64 > PARALLELISM_THRESHOLD {
```
Shnatsel, line 210:
```suggestion
                if width * height > PARALLELISM_THRESHOLD {
```
 The number of channels doesn't seem to matter.

Shnatsel:
Added heuristics based on my quick testing.

PR #252, opened 2022-07-10
Introduces functionality for optionally returning incomplete pixel data

quilan1:
Thought I might start the conversation around a manner in which one could recover partially rendered image pixel data from a corrupt or partially invalid jpeg file.

Still to do / discuss:
- [ ] Add a test image, but I'm not 100% sure where to put it, and/or write a test for it.
- [ ] Decide on the error format: Should only one error be stored? If multiple errors are recoverable, should this be a vector of Errors? Or just the first one?
- [ ] Should the TryRecover trait be externally visible, or rather a new decode function made? (and pull the recovery machinery inside)

Just wanted to throw out a minimally invasive rough draft, perhaps to start some thinking about things.

======================

Currently, this library is rather rigid, with respect to the specification.
If an error occurs while decoding the stream, the error is returned and
recovery of the partially decoded image is impossible.

However, there are some occasions where it may be possible, to return the
potentially incomplete image data. In this case, the `try_recover` function
may be used on the result of the `decode` function. If the data supports
recovery, the incomplete pixels are returned, but if not, it simply returns
the error.

Added the `TryRecover` trait, the `Error::Recoverable` condition and
implemented a simple example usage -- if an image does not have an EOI
segment, it yields a recoverable error.

Usage of the functionality is straightforward:

```
use jpeg_decoder::TryRecover;

let mut decoder = Decoder::new(File::open("missing-eoi.jpg")?);

// Bails with an IO error about UnexpectedEof
// let pixels = decoder.decode()?;

// Returns the incomplete pixel data of the image
let pixels = decoder.decode().try_recover()?;
```

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -272,6 +272,7 @@ impl<R: Read> Decoder<R> {
             ));
         }
 
+        let mut recoverable_error = None;
         let mut previous_marker = Marker::SOI;
         let mut pending_marker = None;
         let mut scans_processed = 0;
@@ -291,7 +292,13 @@ impl<R: Read> Decoder<R> {
         loop {
             let marker = match pending_marker.take() {
                 Some(m) => m,
-                None => self.read_marker()?,
+                None => match self.read_marker() {
+                    Err(e) => {
+                        recoverable_error = Some(e);
+                        break;
+                    },
+                    Ok(v) => v,
+                },
             };
 
             match marker {
@@ -566,10 +573,16 @@ impl<R: Read> Decoder<R> {
         let frame = self.frame.as_ref().unwrap();
         let preference = Self::select_worker(&frame, PreferWorkerKind::Multithreaded);
 
-        worker_scope.get_or_init_worker(
+        let pixels = worker_scope.get_or_init_worker(
             preference,
             |worker| self.decode_planes(worker, planes, planes_u16)
-        )
+        )?;
+
+        if let Some(error) = recoverable_error {
+            Err(Error::Recoverable { err: Box::new(error), pixels: pixels })
+        } else {
+            Ok(pixels)
+        }
     }
 
     fn decode_planes(
--- a/src/error.rs
+++ b/src/error.rs
@@ -30,7 +30,6 @@ pub enum UnsupportedFeature {
 }
 
 /// Errors that can occur while decoding a JPEG image.
-#[derive(Debug)]
 pub enum Error {
     /// The image is not formatted properly. The string contains detailed information about the
     /// error.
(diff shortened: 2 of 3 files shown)


src/error.rs
```diff
@@ -69,3 +88,20 @@ impl From<IoError> for Error {
...
+}
+
+impl TryRecover for Result<Vec<u8>> {
+    fn try_recover(self) -> Self {
+        self.or_else(|err| match err {
+            Error::Recoverable { pixels, .. } => {
+                Ok(pixels)
+            },
+            _ => Err(err),
+        })
```
HeroicKatora, line 105: I would rather provide this as an inherent method of `Error`:

```
impl Error {
    pub fn partial_recover(self) -> Result<Vec<u8>, Error>;
}
```

Each usage site can easily emulate the above trait by calling `Result::{map_,}or_else` themselves which is even slightly more powerful by allowing other `Ok`-types than `Vec<u8>`.

src/decoder.rs
```diff
@@ -291,7 +292,13 @@ impl<R: Read> Decoder<R> {
         loop {
             let marker = match pending_marker.take() {
                 Some(m) => m,
-                None => self.read_marker()?,
+                None => match self.read_marker() {
+                    Err(e) => {
```
HeroicKatora, line 296: Do we really want to ingore all IO errors? see `io::Error::kind()` and `ErrorKind::UnexpectedEof`, I would start with a small set.

src/error.rs
```diff
@@ -41,6 +40,13 @@ pub enum Error {
     Io(IoError),
     /// An internal error occurred while decoding the image.
     Internal(Box<dyn StdError + Send + Sync + 'static>), //TODO: not used, can be removed with the next version bump
+    /// An error that occurred during the decode, but allows for incomplete image pixel data to be returned with `try_recover()`
+    Recoverable {
+        /// Error occurred while decoding the image
+        err: Box<Self>,
+        /// Incomplete pixel data of the image
+        pixels: Vec<u8>
+    }
```
HeroicKatora, line 49: If the variant stored a `struct RecoverableError` that contains the fields instead we could later add different error information and state in a SemVer compatible upgrade. And provide methods specifically for recoverable errors, such as listing of errors or a summary of their kinds / specificity. Nothing specific comes to mind but it's been quite useful to keep the option available.

Compare roughly design of `image::ImageError` for *some inspiration*, maybe, but I don't want to cause rash complexification either. Simply these fields in a struct seems fine to me for the moment. That would answer the question of multiple errors by not having to answer it, yet :slightly_smiling_face:

HeroicKatora:
I've added my thoughts on idiomatic code style above but in terms of functionality, I do believe this is already decently covers a few cases. We clearly signal when a result is not standard compliant (except for implementation bugs) but have an interface to return the main decoding result as far as possible nevertheless.

Finding a non-license encumbered test is always a hassle. However, as a guidance you can add a subfolder `tests/recover` structured similar to `reftest`. I'm not sure which properties we even want to test for these images, though. Maybe start with asserting that a recoverable error is returned without checking the exact contents?

PR #254, opened 2022-07-28
Enable cloning of `UnsupportedFeature` to allow use in image-tiff crate

AlanRace:
To solve https://github.com/image-rs/image-tiff/issues/168 I would like to clone the `UnsupportedFeature` error and add this to a `TiffError` to pass back to the caller.


fintelia (approved):

micahsnyder:
It appears this is approved but hasn't merged. @HeroicKatora does this need your sign-off?

HeroicKatora:
We usually wait for CI but it's not caused by this PR:

>       rust wasm file schema version: 0.2.82
>        this binary schema version: 0.2.80 (4caa98165)

Issue #238, opened 2022-05-02
Provide ability to override colour transform?


AlanRace:
Decompressing JPEG encoded .tiff files causes incorrect colour conversion (e.g. https://github.com/image-rs/image-tiff/issues/124). This has been temporarily resolved by modifying the JPEG header before being passed to `jpeg-decoder` (see https://github.com/image-rs/image-tiff/pull/148). This is not an ideal solution and feels somewhat hacky.

As the tiff header includes information on the `PhotometricInterpretation` and necessary parameters for colour conversion, my ideal solution would be to provide an option to ensure that `jpeg-decoder` doesn't do any form of colour correction/conversion and simply passes back the decompressed data, so that this can be handled by the image-tiff library (https://github.com/image-rs/image-tiff). Does that sound reasonable? If so I can propose a PR.

HeroicKatora:
The approach sounds good, in particular ensuring that `Upsampler::upsample_and_interleave_row` is passed a no-op for its `color_convert`, and a PR would be great.

Issue #249, opened 2022-07-04
JPEG image decoded incorrectly


Shnatsel:
This happens in image v0.24.2

## Expected

![cbsimbanalaor4](https://user-images.githubusercontent.com/291257/177062315-837cd206-e8c8-4f86-9e6b-072274724077.jpg)

## Actual behaviour

![image](https://user-images.githubusercontent.com/291257/177062381-f4683f7b-1f80-449a-bca0-5344f149b975.png)

## Reproduction steps

```rust
use std::error::Error;

fn main() -> Result<(), Box<dyn Error>> {
    use image::io::Reader as ImageReader;
    let input = std::env::args().nth(1).unwrap();
    let output = std::env::args().nth(2).unwrap();
    
    let img = ImageReader::open(input)?.decode()?;
    img.save(output)?;
    Ok(())
}

```

invoked with `image-convert in.jpg out.png`

quilan1:
I might be wildly out there, because it's not my typical wheelhouse, but it seems like the following is occurring:

The application segment has a color transform value of `AdobeColorTransform::Unknown` (0) which, inside `choose_color_convert_func`, we see:

```
match component_count {
    3 => {
        // http://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/JPEG.html#Adobe
        // Unknown means the data is RGB, so we don't need to perform any color conversion on it.
        if color_transform == Some(AdobeColorTransform::Unknown) {
            Ok(color_convert_line_rgb)
        } else {
            Ok(color_convert_line_ycbcr)
        }
    }
```

I think, and I might be getting mixed up about this, that the data's actually in YCbCr, not RGB. In this case, a maxed out pixel YCbCr (1.0, 0.5, 0.5), or (0xFF, 0x80, 0x80) is interpreted as RGB 0xFF8080 -- or the magenta color of the background.

If you remove the segment @ file offset 0x190, the image decodes fine and if you remove the if statement in the code above and just replace it with `Ok(color_convert_line_ycbcr)`, it works fine, but I don't know what the correct approach is to do. I'd love to see other files with the Unknown color transform, to get a better idea of the data out there.

AlanRace:
@ Shnatsel This should now be resolved in the master branch - can you check that this hasn't had any unintended consequences for your large set of JPEGs?

Shnatsel:
I'll see if I can run a before/after test.

Shnatsel:
[...]

PR #255, opened 2022-07-28
Override colour transform

AlanRace:
Added the ability to set the colour transform applied (resolves https://github.com/image-rs/jpeg-decoder/issues/238).

To do this, I had added a new `enum` which captures the possible options, including a `Default` option which retains the previous behaviour when the `color_transform == None`.

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -110,7 +110,8 @@ pub struct Decoder<R> {
     quantization_tables: [Option<Arc<[u16; 64]>>; 4],
 
     restart_interval: u16,
-    color_transform: ColorTransform,
+    fallback_color_transform: ColorTransform,
+    color_transform: Option<ColorTransform>,
     is_jfif: bool,
     is_mjpeg: bool,
 
@@ -137,8 +138,9 @@ impl<R: Read> Decoder<R> {
             ac_huffman_tables: vec![None, None, None, None],
             quantization_tables: [None, None, None, None],
             restart_interval: 0,
-            // Set the default transform as unknown. This can be overriden using `set_color_transform`
-            color_transform: ColorTransform::default(),
+            // Set the fallback transform as unknown. This can be overriden using `set_color_transform`
+            fallback_color_transform: ColorTransform::default(),
+            color_transform: None,
             is_jfif: false,
             is_mjpeg: false,
             icc_markers: Vec::new(),
@@ -149,9 +151,15 @@ impl<R: Read> Decoder<R> {
         }
     }
 
-    /// Sets the colour transform to be applied before returning the data.
+    /// Colour transform to use if one is not set using `set_color_transform` or found in app segments.
+    pub fn set_fallback_color_transform(&mut self, transform: ColorTransform) {
+        self.fallback_color_transform = transform;
+    }
+
+    /// Colour transform to use when decoding the image. App segments relating to colour transforms
+    /// will be ignored.
     pub fn set_color_transform(&mut self, transform: ColorTransform) {
-        self.color_transform = transform;
+        self.color_transform = Some(transform);
     }
 
     /// Set maximum buffer size allowed for decoded images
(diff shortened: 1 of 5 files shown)


src/decoder.rs
```diff
@@ -74,6 +74,31 @@ pub struct ImageInfo {
     pub coding_process: CodingProcess,
 }
 
+/// Describes the colour transform to apply before binary data is returned
+#[derive(Debug, Clone, Copy)]
+pub enum ColorTransform {
```
HeroicKatora, line 79: Please add `#[non_exhaustive]` on here, just in case.

src/decoder.rs
```diff
@@ -492,7 +520,8 @@ impl<R: Read> Decoder<R> {
                     if let Some(data) = parse_app(&mut self.reader, marker)? {
                         match data {
                             AppData::Adobe(color_transform) => {
-                                self.color_transform = Some(color_transform)
+                                self.color_transform =
+                                    ColorTransform::AdobeColorTransform(color_transform)
```
HeroicKatora, line 524: This seems like it leads to confusing semantics. If one explicitly sets up `ColorTransform::None` but then it's overwritten by an app segment? Which semantics did you intend, could you document them on `set_transform` as well?

AlanRace: Hmm this is a good point. What do you think to:

```rust
/// Sets the colour transform to be applied before returning the data. If must_use is set to true, 
/// then this transform will be used for decoding, and app segments relating to colour transforms 
/// will be ignored. If must_use is set to false, then this transform will be used only in the case 
/// where no app segment overwrites it.
    pub fn set_color_transform(&mut self, transform: ColorTransform, must_use: bool)
```

HeroicKatora: Not sure, I'm not too fond of `bool` parameters but would maybe use a separate function for 'hints' vs. overrides, for instance because not all variants may be applicable as hints. Also, a hint might have the semantics of falling back to the default behavior in case `CMYK` is not applicable where an override errors in 3-component color.

Choices, so many choices :smile:

AlanRace: Not quite sure if I understand what you mean with 'hints', but we could do:

```rust
/// Colour transform to use if one is not set using `set_color_transform` or found in app segments.
pub fn set_fallback_color_transform(&mut self, transform: ColorTransform) {
...
}

/// Colour transform to use when decoding the image. App segments relating to colour transforms 
/// will be ignored.
pub fn set_color_transform(&mut self, transform: ColorTransform) {
...
}
```

Alternatively, we could split out the fallback based on number of channels, like so:

```rust
/// Colour transform to use when the data has 3 channels and if one is not set using `set_color_transform` or found in app segments.
pub fn set_fallback_color_transform_3_channels(&mut self, transform: ColorTransform) {
...
}

```rust
/// Colour transform to use when the data has 4 channels and if one is not set using `set_color_transform` or found in app segments.
pub fn set_fallback_color_transform_4_channels(&mut self, transform: ColorTransform) {
...
}

```

But those are some long function names....

HeroicKatora: Option 1 actually looks good.

fintelia: Yeah. Perhaps even `set_fallback_color_transform` vs. `override_color_transform`.

AlanRace:
Hopefully this should now resolve the previous ambiguity.

I decided that the presence of app segments should set `color_transform` rather than overwriting `fallback_color_transform`, but that's a simple change if you prefer it the other way.

AlanRace:
As this pull request is dealing with colour transforms, I had a look at solving https://github.com/image-rs/jpeg-decoder/issues/249 too. I have rewritten the logic for selecting the colour transform based on the information [here](https://entropymine.wordpress.com/2018/10/22/how-is-a-jpeg-images-color-type-determined/).

That (hopefully) removes the need for the fallback transform being set, so I have removed it. Let me know what you think.

fintelia (approved):

PR #264, opened 2022-10-29
Wasm simd

dustletter:
This adds a SIMD implementation using simd-128, along with a small change to fix link errors for the wasm32-wasi target.  It speeds up the large_image benchmark by about 45%.
All tests pass except for large_image.jpg - the image dimensions overflow on 32-bit platforms.
In order to run benchmarks, Criterion needs to be updated to version 0.4, which makes Rayon an optional feature.  I didn't check in this change because I didn't want to mess with dependencies for a mostly unrelated change.

Sorry for making an unsolicited pull request!  I did this for a hobby project and thought it might be worth upstreaming.

### DIFF
--- a/src/arch/wasm.rs
+++ b/src/arch/wasm.rs
@@ -1,0 +1,270 @@
+#[cfg(target_arch = "wasm32")]
+use std::arch::wasm32::*;
+
+#[cfg(target_arch = "wasm32")]
+#[target_feature(enable = "simd128")]
+fn idct8(data: &mut [v128; 8]) {
+    // The fixed-point constants here are obtained by taking the fractional part of the constants
+    // from the non-SIMD implementation and scaling them up by 1<<15. This is because
+    // i16x8_q15mulr_sat(a, b) is effectively equivalent to (a*b)>>15 (except for possibly some
+    // slight differences in rounding).
+
+    // The code here is effectively equivalent to the calls to "kernel" in idct.rs, except that it
+    // doesn't apply any further scaling and fixed point constants have a different precision.
+
+    let p2 = data[2];
+    let p3 = data[6];
+    let p1 = i16x8_q15mulr_sat(i16x8_add_sat(p2, p3), i16x8_splat(17734)); // 0.5411961
+    let t2 = i16x8_sub_sat(
+        i16x8_sub_sat(p1, p3),
+        i16x8_q15mulr_sat(p3, i16x8_splat(27779)), // 0.847759065
+    );
+    let t3 = i16x8_add_sat(p1, i16x8_q15mulr_sat(p2, i16x8_splat(25079))); // 0.765366865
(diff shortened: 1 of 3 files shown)


HeroicKatora:
Very nice. Definitely worth upstreaming! And 45%, that's just incredible :)

HeroicKatora (approved):
I'm definitely no expert in wasm simd but LGTM :+1: It's interesting that the load/store operations have no alignment requirements so we can avoid that stack buffer nonoverlapping copy we have in other implementations.

src/arch/wasm.rs
```diff
@@ -0,0 +1,270 @@
...
+        // increased by 3.
+        // As values will be stored in a u8, they need to be 128-centered and not 0-centered.
+        // We add 128 with the appropriate shift for that purpose.
+        const OFFSET: i16 = 128 << (SHIFT + 3);
+        // We want rounding right shift, so we should add (1/2) << (SHIFT+3) before shifting.
+        const ROUNDING_BIAS: i16 = (1 << (SHIFT + 3)) >> 1;
+
+        let data_with_offset = i16x8_add_sat(data[i], i16x8_splat(OFFSET + ROUNDING_BIAS));
+
+        unsafe {
```
fintelia, line 175: Could you add a `// SAFETY:` comment explaining why this block is OK? Specifically that the assertion above makes sure the wrapping_add doesn't go out of bounds

fintelia (approved):
Also not a wasm simd expert, but the unsafe blocks look good to me at least

dustletter:
> It's interesting that the load/store operations have no alignment requirements so we can avoid that stack buffer nonoverlapping copy we have in other implementations.

This was surprising to me too.  In the spec, it's possible to have a higher alignment, but Rust decided against making it the default. (side note: is it possible/worth it to align those slices for the other versions?)

This was almost all done by copying the SSSE3 implementation and find-and-replacing instructions (e.g. _mm_mulhrs_epi16 to i16x8_q15mulr_sat), with the exception of `transpose()` and the shuffle to RGB order in `color_convert_line_ycbcr()`.  I admit I took the pointer arithmetic for granted 😅   I've gone through and added safety comments to all three unsafe blocks.

mcroomp:
What about just using the Wide crate and letting the compiler do the rest? It does an amazing job picking the right instructions depending on what features you enable (SSE2,AVX,WASM) etc. Even the transpose is done surprisingly fast despite looking like it would be super slow.

Here's an example implementation:
https://github.com/microsoft/lepton_jpeg_rust/blob/main/src/structs/idct.rs

HeroicKatora:
Sorry for forgetting to merge this after approvals.

@mcroomp Feel free to provide a PR with performance comparison but the experience with any auto-vectorization-based approach has, historically, not been amazing.

PR #267, opened 2023-01-04
Collect unknown app1 data for further processing

taavit:
I'm working on project with some non-exif tags data, and only currently only exif tag from App(1) is collected. This PR collects all unknown app1 data, allowing user of library its further processing.


HeroicKatora (commented):
The concern isn't seriously blocking but if you want to provide a more idiomatic iterator API, please do so.

Otherwise, I'd return `Vec<Vec<u8>>` and not worry too much about the additional clone, just like we do for `icc_profile`. Consistency will be more important here than that one allocation (in terms of maintanance anyways).

src/decoder.rs
```diff
@@ -197,6 +199,11 @@ impl<R: Read> Decoder<R> {
         self.exif_data.as_deref()
     }
 
+    /// Returns collection of raw app data that couldn't be recognized
+    pub fn unknown_data(&self) -> &Vec<Vec<u8>> {
```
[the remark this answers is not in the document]
taavit: I've ended with, although I'm not sure if map is a right tool here.
```rust
   /// Returns collection of raw app data that couldn't be recognized
   pub fn unknown_data(&self) -> impl Iterator<Item=&'_ [u8]> + '_ {
       self.unknown_data.iter().map(|v| { v.as_slice() })
   }
```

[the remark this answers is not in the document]
fintelia: I'd personally lead towards `&[Vec<u8>]`. The type signature of the method necessitates that all the raw app data has already been loaded before calling it so there doesn't seem to be much benefit to returning an iterator.

src/decoder.rs
```diff
@@ -197,6 +199,11 @@ impl<R: Read> Decoder<R> {
         self.exif_data.as_deref()
     }
 
+    /// Returns collection of raw app data that couldn't be recognized
+    pub fn unknown_data(&self) -> impl Iterator<Item=&'_ [u8]> + '_ {
```
fintelia, line 203: Calling this `unknown_data` feels a bit vague.

Also, I'm noticing that the "unknown" part of the name signals a possible backwards compatibility issue: if later versions of the crate do recognize some other app data, will the user have to switch interfaces?

taavit: Then maybe store all AppData, including known one. So in the future, when crate recognizes more tags, user will switch from custom implementation of "raw" to crate one?

HeroicKatora: Really makes me consider if data from all `App(n)` variants should be collected?

Just as an overview, there's a whole lot of specific cases that would access these and which we currently do not store: https://exiftool.org/TagNames/JPEG.html

In that case however the `Unknown` variant should collect the `n` so that the caller always knowns which kind of app segment the data is from.

HeroicKatora: What about a simple struct? Should be semantically cleaner than sticking to tuples. And with that, fintelias preference in the other comment makes more sense to me, i.e. the public API can return `&[AppSegment]` for accessing them instead of that impl-iterator type.

```rust
struct AppSegment {
    pub kind: u8,
    pub data: Vec<u8>,
}
```

fintelia: Also, the exif data can just be stored as an index in the `Vec<AppSegment>` rather than needing to keep around an extra copy

taavit: The difference between exif data `[u8]` and app_segment with exif `[u8]` is header `*b"Exif\x00\x00"`.

hfiguiere:
I was looking at adding explicit XMP support, which is an APP1 segment starts with the string `http://ns.adobe.com/xap/1.0/\0`. This would probably conflict with this, but complement it too.

I wonder how we can proceed.

src/parser.rs
```diff
@@ -600,86 +606,66 @@ pub fn parse_com<R: Read>(reader: &mut R) -> Result<Vec<u8>> {
...
+                    _ => return Err(Error::Format("invalid color transform in adobe app segment".to_owned())),
+                };
+
+                result = Some(AppData::Adobe(color_transform));
             }
         },
         _ => {},
     }
 
-    skip_bytes(reader, length - bytes_read)?;
```
hfiguiere, line 683: removing this call cause the follwoing warning:

```
warning: function is never used: `skip_bytes`
   --> src/parser.rs:154:4
    |
154 | fn skip_bytes<R: Read>(reader: &mut R, length: usize) -> Result<()> {
    |    ^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

tests/lib.rs
```diff
@@ -153,3 +153,24 @@ fn read_exif_data() {
     // exif data start as a TIFF header
     assert_eq!(&exif_data[0..8], b"\x49\x49\x2A\x00\x08\x00\x00\x00");
 }
+
+#[test]
+fn read_app_segments() {
+    let path = Path::new("tests")
+        .join("reftest")
+        .join("images")
+        .join("unknown_app1.jpg");
```
hfiguiere, line 162: you could use the existing `ycck.jpg`. It has all the same app1 and more.

taavit:
> I was looking at adding explicit XMP support, which is an APP1 segment starts with the string `http://ns.adobe.com/xap/1.0/\0`. This would probably conflict with this, but complement it too.
> 
> I wonder how we can proceed.

Maybe app0/app1 could be separated from parsing jpeg data, then it could be more generic, and allows custom parser to be injected. Otherwise handling new custom app1 data will need changes to this library, or creating forks.

PR #272, opened 2023-09-25
Admit more precision values in JPEG Lossless

Enet4:
It appears that the implementation of JPEG lossless originally provided in #193 can also handle sample precision values other than 8 and 16. The standard itself covers the lossless coding between 2 and 16 bits of sample precision. This PR relaxes the frame precision check to let the algorithms proceed in this particular case. I also added some public test files with 10 bits and 12 bits of precision for validation.

### Summary

- Relocate the previous lossless JPEG test file set
- Relax the checks to accept precision values in `2..=16` when coding process is lossless (continues to accept only 8 bits otherwise)
- Map newly admitted precision values to the right pixel format when retrieving image meta information
- Add a new set of reftest files from DICOM-WG04 data set compilation


Enet4:
If there is something I can do to help this get merged, please let me know. JPEG lossless is not supported by any other JPEG decoder written in Rust at the moment, and this feature is particularly important for the DICOM-rs project.

fintelia:
Sorry for the delay on this!

Enet4:
Hello again! There hasn't been a crate release for this and other changes made in 2023. Is there a chance that this could be done eventually? Much appreciated! 🙏 

In the meantime, I have already reached out to zune-jpeg to see how to incorporate JPEG lossless decoding there (https://github.com/etemesi254/zune-image/issues/144), but it might take some time before I or anyone else can pick this up.

PR #274, opened 2023-10-04
Use core and alloc a bit more

linkmauve:
We still depend mostly on `std::io` and `std::fs`, but the rest can be done without std.

### DIFF
--- a/src/arch/ssse3.rs
+++ b/src/arch/ssse3.rs
@@ -1,7 +1,8 @@
+use core::ptr;
 #[cfg(target_arch = "x86")]
-use std::arch::x86::*;
+use core::arch::x86::*;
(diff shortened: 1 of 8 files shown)


fintelia:
Could you say a bit about the motivation for this PR? The remaining usages of `std` are pretty fundamental to the crate, so I don't think we'll likely be able to switch to `nostd` any time soon (and at some point zune-jpeg will probably replace most use of this crate)

linkmauve:
I’m upstreaming some of the changes I had to do to run this JPEG decoder on a GameCube in `no_std`. The rest of the changes are somewhat similar to #226, except I didn’t attempt to stay compatible with `std::io`.

PR #279, opened 2024-10-30
Format Rust code

jayvdb:
Previous PR https://github.com/image-rs/jpeg-decoder/pull/271 over a year old and has no feedback.

Older open PRs, also no feedback
- https://github.com/image-rs/jpeg-decoder/pull/248
- https://github.com/image-rs/jpeg-decoder/pull/154

I didnt see any closed PRs.

### DIFF
--- a/benches/decoding_benchmark.rs
+++ b/benches/decoding_benchmark.rs
@@ -18,24 +18,28 @@ fn read_metadata(image: &[u8]) -> ImageInfo {
 
 fn main() {
     let mut c = Criterion::default().configure_from_args();
-    c.bench_function("decode a 512x512 JPEG", |b| b.iter(|| {
-        read_image(include_bytes!("tower.jpg"))
-    }));
+    c.bench_function("decode a 512x512 JPEG", |b| {
+        b.iter(|| read_image(include_bytes!("tower.jpg")))
+    });
 
-    c.bench_function("decode a 512x512 progressive JPEG", |b| b.iter(|| {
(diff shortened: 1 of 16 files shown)


fintelia:
This project is in maintenance mode and not seeking further development. Additionally, reformatting PRs like this can really only be done by maintainers because manually reviewing every change in a PR like this would be both time consuming and pointless

jayvdb:
We rely on the lossless functionality and have related fixes to submit. c.f. https://github.com/franklin-ai/jpeg-decoder/commits/main/

If you would like to re-do this PR yourself, here are the steps that I took

1. added `#[rustfmt::skip]` before `pub fn fill_default_mjpeg_tables(` and the nicely formatted tables at https://github.com/image-rs/jpeg-decoder/blob/4035b7e/src/arch/wasm.rs#L245-L261 , so that rustfmt doesnt make a mess of them.
2. In `src/huffman.rs`, `src/parser.rs`, `src/upsampler.rs` , `src/worker/immediate.rs`, `tests/lib.rs`, and `tests/rayon*.rs` , add a blank line between std, external and crate `use` statements, otherwise `rustfmt` (stable) will jumble them all up.
3. Run `cargo fmt`

That is all.

kornelski:
This isn't an improvement. Every rustfmt change is making the code slightly worse, and peppers it with `#[rustfmt::skip]` unnecessarily.

If you have an editor that unconditionally reformats, then `.rustfmt.toml` with `disable_all_formatting = true` can help.

PR #282, opened 2025-03-05
Fix panic when prediction calculation is invalid

richard-viney:
This bug was found during fuzz testing of a library that depends on `jpeg-decoder`, on the following line:

https://github.com/image-rs/jpeg-decoder/blob/e70564bbff4f9fefcee72efad5874fddd0d6a83d/src/decoder/lossless.rs#L112

It's triggered when the math underflows in the situation that `scan.point_transform` isn't less than `frame.precision`.

This PR adds a check in the parser that errors when this situation is detected, as well as the test case generated by the fuzz tester.

(I also changed to explicitly specify Ubuntu 22.04 in CI due to failures running on 24.04).


.github/workflows/rust.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
fintelia, line 11: I saw that the cross_compile jobs failed due to Ubuntu version, but do you know whether this one can stay Ubuntu latest?

fintelia (approved):

.github/workflows/rust.yml
richard-viney: I'll see if the issue on 24.04 can be fixed, may just be a package that changed its name

richard-viney: Fixed!

fintelia:
Thanks!

PR #287, opened 2026-02-16
Fix lossless restart boundary handling in reconstruction

### LINKED ISSUE #286: JPEG Lossless restart boundaries not applied during predictor reconstruction
### Summary
JPEG Lossless images with restart intervals (`DRI` + `RSTn`) can decode to visibly corrupted output (banding/striping) while decoding succeeds without error.

### Suspected root cause
In `decode_scan_lossless` (`src/decoder/lossless.rs`), restart handling is tracked during entropy decoding, but reconstruction later does not reliably apply restart resets at the same sample boundaries.

In practice this means the `restart` flag used by prediction can be wrong during reconstruction because restart-boundary context is no longer aligned with sample index when reconstructing the `differences` buffer.

### Notes
- This is reproducible on valid JPEG Lossless streams using restart markers.
- A fix should ensure restart resets are derived from absolute sample position (or equivalent preserved state) during reconstruction.

### Proposed patch (minimal)
The core change is to derive restart boundaries from absolute sample index during reconstruction, instead of reusing transient decoder loop state.

```diff
diff --git a/src/decoder/lossless.rs b/src/decoder/lossless.rs
@@
- let prediction = predict(..., self.restart_interval > 0
-     && mcus_left_until_restart == self.restart_interval - 1);
+ let pixel_index = mcu_y * width + mcu_x;
+ let prediction = predict(..., self.restart_interval > 0
+     && pixel_index > 0
+     && pixel_index % self.restart_interval as usize == 0);
```

And in the `Predictor::Ra` branch, apply the same restart condition (`pixel_index % restart_interval == 0`) when selecting the reset predictor at restart boundaries.

### Examples (Before / After)

Image
Image

TidalPaladin:
## Motivation
Lossless JPEG scans with restart markers can decode incorrect pixel values if prediction is not reset exactly at restart boundaries.

## Solution
Detect restart boundaries from the absolute pixel index and reset predictor initialization at those boundaries. Keep the reconstruction logic aligned across predictor paths with shared indexing helpers.

## Changes
- Fix restart-boundary handling in lossless reconstruction.
- Apply the same boundary behavior in both `Predictor::Ra` and generic predictor paths.
- Add a regression reftest fixture (`.jpg` + expected `.png`) that exercises lossless restart-marker decoding.

Closes #286

Generated with Codex


jasonklotzer:
We independently hit this exact bug (before finding this issue/PR) while debugging a rendering failure in [dcmnorm](https://github.com/pohcee/dcmnorm), a DICOM toolkit that uses `jpeg-decoder` via `dicom-transfer-syntax-registry`. A real customer mammography DICOM file (Hologic tomosynthesis/C-View, transfer syntax `1.2.840.10008.1.2.4.70`) failed to render correctly, showing periodic horizontal banding while decoding "succeeded" — same symptom described in #286.

That file has `DRI=63872` samples, which is exactly 32 image rows (1996 × 32), predictor selection Ra (PSV=1). That's the worst case for this bug: every restart boundary falls exactly on column 0 of a row, so the `Predictor::Ra` fast path's "start of line, use pixel above" logic silently absorbs the restart reset instead of applying the default prediction. We confirmed the corrupted output numerically — flat background bands decreased by exactly 512 (8 in the real 10-bit sample domain) every single restart interval, a clean signature of the reconstructed baseline drifting forward across restarts instead of resetting.

We wrote an equivalent fix independently (tracking restart-marker positions and using them in reconstruction for both the `Ra` fast path and the generic predictor path, which has the same bug via the stale `mcus_left_until_restart` check) before finding this PR. The diagnosis and approach here match ours exactly. Applying this patch resolves the corruption completely on our real-world file — verified visually against the file (and against another viewer's rendering of the same study) and against dcmnorm's full test suite (74 tests, no regressions).

One thing that might be blocking merge: the `cross_compile_aarch64` jobs are failing on this PR's CI run, while they pass on current `master`. Everything else (all `build` matrix jobs across stable/beta/nightly, wasm) is green. Given this change only touches `src/decoder/lossless.rs` and doesn't touch the NEON/aarch64-specific IDCT code, it's possible that's a transient/flaky failure rather than something this diff caused — might just need a re-run.

Given restart intervals are common in mammography JPEG Lossless (at least Hologic uses them), this affects real clinical images beyond synthetic test cases. Would be great to see this merged.

